PR #15, opened 2018-06-23
npm translator: Fix endless loop in case of dependency cycles

codeworrior:
Fixes https://github.com/SAP/ui5-project/issues/14 .

### DIFF
--- a/lib/translators/npm.js
+++ b/lib/translators/npm.js
@@ -235,6 +235,7 @@ class NpmTranslator {
 
 	processPendingDeps(tree) {
 		const queue = [tree];
+		const visited = new Set();
 
 		// Breadth-first search to prefer projects closer to root
 		while (queue.length) {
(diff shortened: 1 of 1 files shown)


lib/translators/npm.js
```diff
@@ -235,13 +235,18 @@ class NpmTranslator {
...
 		const queue = [tree];
+		const visited = new Set();
 
 		// Breadth-first search to prefer projects closer to root
 		while (queue.length) {
 			const project = queue.shift(); // Get and remove first entry from queue
 			if (!project.id) {
 				throw new Error("Encountered project with missing id");
 			}
+			if (visited.has(project.id)) {
```
codeworrior: Would it make sense to to check for the combination of `project.id` and `project.version` instead of only the `project.id`?

RandomByte: The projectPreprocessor does the same also on id-level: [projectPreprocessor.js#L55](https://github.com/SAP/ui5-project/blob/master/lib/projectPreprocessor.js#L55). So I guess it wouldn't change much down that line.

While it is yet to be decided and documented, we once agreed (for the time being) to always prefer the projects closer to the root, ignoring the version.

codeworrior:
How _could_ I miss that...

PR #62, opened 2018-11-15
[FEATURE] projectPreprocessor: Allow application project dependency on non-root level

### LINKED ISSUE #61: Server is not considering application dependency
## Expected Behavior
If you are developing on a library (or future component types see related issue https://github.com/SAP/ui5-tooling/issues/45) and you add an application project as a 
`dependency` or `devDependency`, the application dependency should be served as the server root.

Also the dependent applications index.html etc. should be served.

## Current Behavior
Currently the application dependency is not served at all when beeing added as a dependency.

The application appears in ui5 tree, but it'll still not be served.

## Steps to reproduce the issue
1. Create a library project
2. Add a application project as dependency
3. Try to serve

## Affected components
- [ ] [ui5-builder](https://github.com/SAP/ui5-builder)
- [x] [ui5-server](https://github.com/SAP/ui5-server)
- [ ] [ui5-cli](https://github.com/SAP/ui5-cli)
- [ ] [ui5-fs](https://github.com/SAP/ui5-fs)
- [x] [ui5-project](https://github.com/SAP/ui5-project)
- [ ] [ui5-logger](https://github.com/SAP/ui5-logger)

RandomByte:
Resolves #61

### DIFF
--- a/lib/projectPreprocessor.js
+++ b/lib/projectPreprocessor.js
@@ -283,12 +283,19 @@ class ProjectPreprocessor {
 			throw new Error(`Root project needs to be of kind "project". ${project.id} is of kind ${project.kind}`);
 		}
 
-		if (project.kind === "project" && project.type === "application" && project._level !== 0) {
-			// There is only one project project of type application allowed
-			// That project needs to be the root project
-			log.verbose(`[Warn] Ignoring project ${project.id} with type application`+
-					` (distance to root: ${project._level}). Type application is only allowed for the root project`);
-			return false; // ignore this project
+		if (project.kind === "project" && project.type === "application") {
+			// There must be exactly one application project per dependency tree
+			// If multiple are found, all but the one closest to the root are rejected (ignored)
+			// If there are two projects equally close to the root, an error is being thrown
+			if (!this.qualifiedApplicationProject) {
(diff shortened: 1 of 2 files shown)


lib/projectPreprocessor.js
```diff
@@ -283,12 +283,19 @@ class ProjectPreprocessor {
...
+		if (project.kind === "project" && project.type === "application") {
+			// There must be exactly one application project per dependency tree
+			// If multiple are found, all but the one closest to the root are rejected (ignored)
+			// If there are two projects equally close to the root, an error is being thrown
+			if (!this.qualifiedApplicationProject) {
+				this.qualifiedApplicationProject = project;
+			} else if (this.qualifiedApplicationProject._level === project._level) {
+				throw new Error(`Found at least two projects ${this.qualifiedApplicationProject.id} and ` +
+					`${project.id} of type application with the same distance to the root project. ` +
+					"Only one project of type application can be used. Failed to decide which one to ignore.");
```
matz3: `Failed to decide which one to ignore.` how to decide this?

RandomByte: The one closer to the root. But if both have the same distance... 🧐

PR #68, opened 2018-11-20
[INTERNAL] npm t8r: Handle projects with explicitly zero UI5 dependencies

RandomByte:


lib/translators/npm.js
```diff
@@ -62,7 +62,7 @@ class NpmTranslator {
 		}
 
 		const ui5Deps = pkg.ui5 && pkg.ui5.dependencies;
-		if (ui5Deps && ui5Deps.length) {
+		if (ui5Deps && ui5Deps instanceof Array) {
```
matz3: Although in node is shouldn't be a problem, I would prefer using isArray:
```suggestion
		if (ui5Deps && Array.isArray(ui5Deps)) {
```

PR #75, opened 2018-12-11
[INTERNAL] static translator and normalizer: added tests

romaniam:


lib/translators/static.js
```diff
@@ -17,13 +17,15 @@ const parseYaml = require("js-yaml").safeLoad;
...
-function generateDependencyTree(dirPath, translatorOptions) {
+function generateDependencyTree(dirPath, translatorOptions = []) {
 	const depFilePath = translatorOptions[0] || path.join(dirPath, "projectDependencies.yaml");
 
 	return new Promise(function(resolve, reject) {
 		fs.readFile(depFilePath, function(err, buffer) {
 			if (err) {
-				reject(err);
+				reject(new Error(
+					`[static translator] Failed to locate projectDependencies.json for "${dirPath}"`)
```
RandomByte: There are all kinds of low-level file system errors that might happen here. Therefore the actual fs-error should be part of the new error message.

Example:
https://github.com/SAP/ui5-project/blob/97e96f3677050901ba097b3fadaea07e95914019/lib/translators/npm.js#L231-L232

RandomByte: I think it should read
> `Failed to locate projectDependencies.yaml at path "${dirPath}"`

rather than
> `Failed to locate projectDependencies.json for "${dirPath}"`

romaniam: Nice, thanks for great feedback. Should be fine now.

lib/normalizer.js
```diff
@@ -16,8 +16,8 @@ const projectPreprocessor = require("./projectPreprocessor");
  * @param {string} options.translator Translator to use
  * @returns {Promise} Promise resolving to tree object
  */
-async function generateProjectTree(options) {
-	const tree = await generateDependencyTree(options);
+async function generateProjectTree(options = {}) {
+	const tree = await this.generateDependencyTree(options);
```
matz3: This is done to be able to mock the function, right?
I don't think that it's good practice though as they are just local functions and at the end, we assign both to the module.exports object. In case `generateProjectTree ` is called as local function, this would break.

RandomByte: Good point, I missed that.

romaniam: > This is done to be able to mock the function, right?
> I don't think that it's good practice though as they are just local functions and at the end, we assign both to the module.exports object. In case `generateProjectTree ` is called as local function, this would break.

I have added a fix 87e9873 for this by wrapping the functions into a module. We will lose the facade in this case but will be able to test the module. Please check if the fix is sufficient

matz3: Yep, that looks much better now 👍

RandomByte:
> COVERAGE INCREASED (+6.4%) TO 80.203%

Nice 🥇

PR #134, opened 2019-02-28
[INTERNAL] Clarify error message for invalid or unsupported specVersions

RandomByte:
Old CLI releases might simply not support a certain specVersion

### DIFF
--- a/lib/projectPreprocessor.js
+++ b/lib/projectPreprocessor.js
@@ -268,8 +268,9 @@ class ProjectPreprocessor {
 
 		if (project.specVersion !== "0.1" && project.specVersion !== "1.0") {
 			throw new Error(
-				`Invalid specification version defined for project ${project.id}: ${project.specVersion}. ` +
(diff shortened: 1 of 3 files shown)


lib/projectPreprocessor.js
```diff
@@ -325,8 +326,8 @@ class ProjectPreprocessor {
 			throw new Error(`No specification version defined for extension ${extension.metadata.name}`);
 		} else if (extension.specVersion !== "0.1" && extension.specVersion !== "1.0") {
 			throw new Error(
-				`Invalid specification version defined for extension ` +
-				`${extension.metadata.name}: ${extension.specVersion}. ` +
+				`Invalid or unsupported specification version ${extension.specVersion} defined for extension ` +
```
matz3: I would even put `unsupported` in front as this is probably the general case.
Or even just replace `Invalid` with `Unsupported` and don't mention `invalid` at all?

RandomByte: Done

PR #180, opened 2019-05-28
[INTERNAL] Configuration docs: Add builder resource excludes

RandomByte:
Documents https://github.com/SAP/ui5-builder/pull/254

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -42,6 +42,8 @@ resources:
     paths:
       "<virtual path 1>": "<physical path 1>"
       "<virtual path 2>": "<physical path 2>"
+  excludes:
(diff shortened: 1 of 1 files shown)


docs/Configuration.md
```diff
@@ -52,6 +52,11 @@ builder:
       afterTask: custom-task-name-1
       configuration:
         color: blue
+  resources:
+    excludes:
+      - /resources/some/project/name/test_results/**
+      - /test-resources/**
+      - !/test-resources/some/project/name/demo-app/**
```
matz3: According to the syntax highlighting, this is not a valid string.
When parsing it, the JSON content is `null`. Quotes should be used.

RandomByte: Done 👍

PR #189, opened 2019-06-24
[FIX] projectPreprocessor: Do not remove already removed dependencies

RandomByte:
This lead to the wrong dependency being removed in some cases.

Credits to @nlunets

### DIFF
--- a/lib/projectPreprocessor.js
+++ b/lib/projectPreprocessor.js
@@ -163,7 +163,7 @@ class ProjectPreprocessor {
 		if (processedProject) {
 			if (processedProject.ignored) {
 				log.verbose(`Dependency of project ${parent.id}, "${project.id}" is flagged as ignored.`);
-				if (parent.dependencies.indexOf(project) !== -1) {
+				if (parent.dependencies.includes(project)) {
 					parent.dependencies.splice(parent.dependencies.indexOf(project), 1);
 				}
 				return true;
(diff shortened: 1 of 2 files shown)


lib/projectPreprocessor.js
```diff
@@ -163,7 +163,9 @@ class ProjectPreprocessor {
 		if (processedProject) {
 			if (processedProject.ignored) {
 				log.verbose(`Dependency of project ${parent.id}, "${project.id}" is flagged as ignored.`);
-				parent.dependencies.splice(parent.dependencies.indexOf(project), 1);
+				if (parent.dependencies.indexOf(project) !== -1) {
```
matz3: You might use Array#includes instead
```suggestion
				if (parent.dependencies.includes(project)) {
```

RandomByte:
Note: **Squash it**

lib/projectPreprocessor.js
```diff
@@ -163,7 +163,9 @@ class ProjectPreprocessor {
 		if (processedProject) {
 			if (processedProject.ignored) {
 				log.verbose(`Dependency of project ${parent.id}, "${project.id}" is flagged as ignored.`);
-				parent.dependencies.splice(parent.dependencies.indexOf(project), 1);
```
Thodd: Oh, splicing index -1 is a nice bug :D

PR #236, opened 2019-11-04
[INTERNAL] Resolve middleware path

nlunets:
Using path.join will make local middleware resolution fail (at least for the static translator), as the join drops the "./" so "./lib/myTask.js" becomes "/lib/myTask.js" which will not be resolved.

@matz3 @RandomByte please have a look :)

## Pull Request Checklist
- [x] Reviewed the [Contributing Guidelines](https://github.com/SAP/ui5-tooling/blob/master/CONTRIBUTING.md#-contributing-code)
    + Especially the [How to Contribute](https://github.com/SAP/ui5-tooling/blob/master/CONTRIBUTING.md#how-to-contribute) section 
- [x] [No merge commits](https://github.com/SAP/ui5-tooling/blob/master/docs/Guidelines.md#no-merge-commits)
- [x] [Correct commit message style](https://github.com/SAP/ui5-tooling/blob/master/docs/Guidelines.md#commit-message-style)

### DIFF
--- a/lib/projectPreprocessor.js
+++ b/lib/projectPreprocessor.js
@@ -496,7 +496,7 @@ class ProjectPreprocessor {
 		}
 		const taskRepository = require("@ui5/builder").tasks.taskRepository;
 
-		const taskPath = path.join(extension.path, extension.task.path);
+		const taskPath = path.resolve(extension.path, extension.task.path);
 		const task = require(taskPath); // throws if not found
 
 		taskRepository.addTask(extension.metadata.name, task);
@@ -511,7 +511,7 @@ class ProjectPreprocessor {
 		}
 		const {middlewareRepository} = require("@ui5/server");
 
-		const middlewarePath = path.join(extension.path, extension.middleware.path);
+		const middlewarePath = path.resolve(extension.path, extension.middleware.path);
 		middlewareRepository.addMiddleware(extension.metadata.name, middlewarePath);
 	}
 }


RandomByte:
Hey @nlunets! Just for better understanding, why can't you use `lib/myTask.js` instead of `./lib/myTask.js`?

nlunets:
Using `lib/myTask.js` `path.join('./', 'lib/myTask.js')` will resolve to `lib/myTask.js`which is an absolute path and require will then fail

RandomByte:
I see, so the project path that you supply via the static translator is not absolute?

nlunets:
In this case it's even the path of a local middleware actually.

nlunets:
To be more clear, the dependencies in the `projectDependencies.yaml` are usually absolute (although they could be relative, that makes especially sense for libraries but there it's actually working fine).
But as soon as I use the static translator even the extension defined in the main ui5.yaml are not resolved properly using relative path.
And in general I would say `path.resolve` is probably a better fit for what you want to do rather than `path.join`

RandomByte:
Ok, fair enough. We even [documented](https://sap.github.io/ui5-tooling/pages/Project/#static-translator) the static translator with relative paths in mind. So the use case is clear.

I'm just wondering whether there might be similar issues further down in the tooling, since the prominent npm translator always provides absolute paths for the projects.

To make this more bullet proof we could have the static translator to resolve all `path` attributes in the given tree upfront. Would that work for you as well?

nlunets:
Yes it should work as well I think.
If you've got a change for me to validate let me know

RandomByte:
I'll look into it.

RandomByte:
I created https://github.com/SAP/ui5-project/pull/237. Could you give that one a try?

RandomByte:
Closing in favor of https://github.com/SAP/ui5-project/pull/237

PR #252, opened 2020-01-10
[FEATURE] Add specification version 1.1

RandomByte:
The new 'theme-library' type requires spec version 1.1:
https://github.com/SAP/ui5-builder/pull/285

Also change wording of the error message for unsupported spec versions
and fix the referenced URL. This should make it clear that the used
version of the UI5 CLI might be outdated.

### DIFF
--- a/lib/projectPreprocessor.js
+++ b/lib/projectPreprocessor.js
@@ -272,11 +272,11 @@ class ProjectPreprocessor {
 			return false; // ignore this project
 		}
 
-		if (project.specVersion !== "0.1" && project.specVersion !== "1.0") {
+		if (project.specVersion !== "0.1" && project.specVersion !== "1.0" && project.specVersion !== "1.1") {
 			throw new Error(
 				`Unsupported specification version ${project.specVersion} defined for project ` +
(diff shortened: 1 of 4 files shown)


lib/projectPreprocessor.js
```diff
@@ -272,11 +272,11 @@ class ProjectPreprocessor {
 			return false; // ignore this project
 		}
 
-		if (project.specVersion !== "0.1" && project.specVersion !== "1.0") {
+		if (project.specVersion !== "0.1" && project.specVersion !== "1.0" && project.specVersion !== "1.1") {
 			throw new Error(
 				`Unsupported specification version ${project.specVersion} defined for project ` +
-				`${project.id}. ` +
-				`See https://github.com/SAP/ui5-project/blob/master/docs/Configuration.md#specification-versions`);
```
matz3: Maybe also adopt the old Configuration.md file to point to the #specifiction-version anchor? Currently it just points to the Configuration page.

RandomByte: Will do 👍

PR #257, opened 2020-01-20
[INTERNAL] Add specification version 2.0a

RandomByte:
Does not add any features but allows projects to already define a
"framework" configuration for use with UI5 CLI 2.0 while staying
compatible with Specification Version 1.1 and already available
UI5 CLI versions.


RandomByte:
As discussed, an alpha spec version is currently not required.

PR #268, opened 2020-03-05
[FEATURE] projectPreprocessor: Log warning when using a deprecated or restricted dependency

RandomByte:
Projects in the tree that are not the root project may trigger a warning
log if they configure any of the following metadata flags:
- deprecated: true
- sapInternal: true

The latter may not trigger a warning if its parent project defines a
metadata flag "allowSapInternal: true".

### DIFF
--- a/test/lib/projectPreprocessor.js
+++ b/test/lib/projectPreprocessor.js
@@ -1,6 +1,8 @@
 const test = require("ava");
+const sinon = require("sinon");
+const mock = require("mock-require");
 const path = require("path");
-const projectPreprocessor = require("../..").projectPreprocessor;
+const projectPreprocessor = require("../../lib/projectPreprocessor");
 const applicationAPath = path.join(__dirname, "..", "fixtures", "application.a");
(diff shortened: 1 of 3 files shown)


test/lib/projectPreprocessor.js
```diff
@@ -1770,3 +1776,147 @@ test("isBeingProcessed: Deduped project is being ignored", (t) => {
...
+	};
+
+	// one warning should be logged for deprecated dependency
+	await preprocessor.checkProjectMetadata(project1, project2);
+
+	t.is(logWarnSpy.callCount, 1, "One warning got logged");
+	t.deepEqual(logWarnSpy.getCall(0).args[0],
+		"Dependency my.project is deprecated and should not be used for new projects!",
+		"Logged expected warning message");
+	mock.stop("@ui5/logger");
```
matz3: You might also use `mock.stopAll()` within `test.afterEach.always` to ensure a proper cleanup

RandomByte: Done

PR #277, opened 2020-03-19
[FEATURE] Support for spec version 2.0

svbender:
BREAKING CHANGE:
Support for spec version 2.0 in
ui5Framework translator and projectPreprocessor.

### DIFF
--- a/lib/translators/ui5Framework.js
+++ b/lib/translators/ui5Framework.js
@@ -71,16 +71,23 @@ const utils = {
 			// Ignoring UI5 Framework libraries in dependencies
 			return ui5Dependencies;
 		}
-		if (project.specVersion === "2.0" && project.framework && project.framework.libraries) {
-			project.framework.libraries.forEach((dependency) => {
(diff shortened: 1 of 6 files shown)


RandomByte:
Please also add new tests for specVersion 2.0 in `test/lib/extension.js` and `test/lib/projectPreprocessor.js` as done here: https://github.com/SAP/ui5-project/pull/252/files

lib/translators/ui5Framework.js
```diff
@@ -71,7 +71,7 @@ const utils = {
 			// Ignoring UI5 Framework libraries in dependencies
 			return ui5Dependencies;
 		}
-		if (project.framework && project.framework.libraries) {
+		if (project.specVersion === "2.0" && project.framework && project.framework.libraries) {
```
RandomByte: The message logged in the else block of this if is not completely correct anymore. Maybe we can log something different if the project defines an incorrect specVersion?

RandomByte:
This is actually not a breaking change but a [FEATURE]. Either change the commit message in a rebase/squash and force push or remember to change it when squash and merging this PR

PR #288, opened 2020-03-25
[INTERNAL] Adopt to task- and middlewareRepository refactorings

RandomByte:
Depends on:
- taskRepository refactoring: https://github.com/SAP/ui5-builder/pull/438
- middlewareRepository refactoring: https://github.com/SAP/ui5-server/pull/302

Also implements:
- Lazy index.js requires

### DIFF
--- a/lib/projectPreprocessor.js
+++ b/lib/projectPreprocessor.js
@@ -587,9 +587,12 @@ class ProjectPreprocessor {
 		const taskRepository = require("@ui5/builder").tasks.taskRepository;
 
 		const taskPath = path.join(extension.path, extension.task.path);
-		const task = require(taskPath); // throws if not found
 
-		taskRepository.addTask(extension.metadata.name, task);
+		taskRepository.addTask({
+			name: extension.metadata.name,
+			specVersion: extension.specVersion,
+			taskPath,
+		});
 	}
 
 	handleServerMiddleware(extension) {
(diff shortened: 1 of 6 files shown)


lib/projectPreprocessor.js
```diff
@@ -587,9 +587,12 @@ class ProjectPreprocessor {
 		const taskRepository = require("@ui5/builder").tasks.taskRepository;
 
 		const taskPath = path.join(extension.path, extension.task.path);
```
matz3: `extension.path` is already absolute, right?

test/lib/extensions.js
```diff
@@ -616,7 +616,7 @@ test("Project with task extension dependency", (t) => {
 	});
 });
 
-test("Project with task extension dependency - task module not found", async (t) => {
+test("Project with task extension dependency - does not throw for invalid task path", async (t) => {
```
matz3: As already mentioned in the other PR, what about a test when the `require` of the custom task fails? Which error does the user get?

lib/projectPreprocessor.js
RandomByte: Yes

test/lib/extensions.js
RandomByte: No error here. The require now happens when a different module (e.g. MiddlewareManager or *Builder) tries to retrieve the middleware or task. I added additional error handling there and some tests.

PR #290, opened 2020-03-26
[INTERNAL] ui5framework resolvers: Add resolveVersion

matz3:


RandomByte:
Windows tests are failing: [`Installer should be called with expected arguments`](https://dev.azure.com/sap/opensource/_build/results?buildId=48056&view=logs&j=30359001-1ff5-5194-39e8-03473bfa86a9&t=74e88f2c-e58c-54a1-690a-44c7f67226fa&l=672)

test/lib/ui5framework/AbstractResolver.js
```diff
@@ -15,7 +21,7 @@ test("AbstractResolver: abstract constructor should throw", async (t) => {
 	}, `Class 'AbstractResolver' is abstract`);
 });
 
-test("AbstractResolver: constructor", (t) => {
+test.serial("AbstractResolver: constructor", (t) => {
```
RandomByte: Just fyi, for debugging/checking logs, there's also a `--serial` flag for AVA: https://github.com/SAP/ui5-project/blob/7c2033c1a5ab839d1aa3c7ebbd999e2b7cfbf3ef/package.json#L25

lib/ui5Framework/AbstractResolver.js
```diff
@@ -147,10 +150,38 @@ class AbstractResolver {
...
+	 * @param {Object} options
+	 */
+	static async resolveVersion(version, {ui5HomeDir, cwd} = {}) {
+		let spec;
+		if (version === "latest") {
+			spec = "*";
+		} else if (versionRegExp.test(version)) {
+			spec = version;
+		} else {
+			throw new Error(`Framework version specifier "${version}" is incorrect or not supported`);
```
RandomByte:
```suggestion
			throw new Error(`Framework version specifier "${version}" is incorrect or not supported. Try "1.76.0" or "latest"`);
```

matz3: Hm, 1.76 will become outdated very soon, so I think that's not a good idea. Maybe we should point to the versionoverview.html instead?

RandomByte: I'm more concerned about *how* the version needs to be specified. But you are right, mentioning an old version here might be irritating.

lib/ui5Framework/AbstractResolver.js
```diff
@@ -147,10 +150,38 @@ class AbstractResolver {
...
+		}
+		const versions = await this.fetchVersions({ui5HomeDir, cwd});
```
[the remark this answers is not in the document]
matz3: Done

lib/ui5Framework/AbstractResolver.js
matz3: Maybe something like `Try "1.xx", "1.xx.x" or "latest"` ?

PR #351, opened 2020-08-31
[INTERNAL] npm translator: Improve detection of collection projects

matz3:
A collection must have an object property "collection" with a nested
object "modules" in its package.json.
Other types or structures should be ignored and the project should not
be handled as collection.

This might happen when other projects define a "collection" property
that is meant to be used for a different purpose
(e.g. https://stenciljs.com/docs/publishing#package-json).


lib/translators/npm.js
```diff
@@ -82,7 +82,8 @@ class NpmTranslator {
 			on projects.
 			Although we could add the collections dependencies as project dependencies to the related modules
 		*/
-		if (!pkg.collection) {
+		const isCollection = typeof pkg.collection === "object";
```
RandomByte, line 85: As discussed, I think we can even check for the `pkg.collection.modules` object being present

matz3: Done

RandomByte (approved):

PR #373, opened 2020-12-01
[INTERNAL] Properly escape path for RegExp in test

matz3:


test/lib/translators/static.js
```diff
@@ -18,13 +19,10 @@ test("Error: Throws if projectDependencies.yaml was not found", async (t) => {
 	const fsStub = sinon.stub(fs, "readFile");
 	fsStub.callsArgWith(1, fsError);
 	const error = await t.throwsAsync(staticTranslator.generateDependencyTree(projectPath));
-	const escapedYamlPath = path.join(projectPath, "projectDependencies.yaml")
-		.replace("\\", "\\\\")
-		.replace("/", "\\/")
-		.replace(".", "\\.");
+	const escapedYamlPath = path.join(projectPath, "projectDependencies.yaml");
```
RandomByte, line 22: `escapedYamlPath` is not escaped anymore, right?

RandomByte (changes requested):

test/lib/translators/static.js
matz3: Right 😄

RandomByte (approved):

PR #382, opened 2021-01-14
[FIX] ui5Framework.Installer: Detect and resolve incomplete installs

RandomByte:
At the beginning of every install, create a marker file unique for that
package and version. If subsequent ui5Framework translator executions
detect such a marker for a package *and* can aquire a lock for it, this
might indicate an aborted or failed install of that package.

In that case, remove the package (if present) and continue with the
normal install process.

Resolves https://github.com/SAP/ui5-tooling/issues/478


RandomByte:
Closing in favor of https://github.com/SAP/ui5-project/pull/389

PR #390, opened 2021-01-28
[FIX] ui5Framework.Installer: Ensure target directory does not exist before rename

RandomByte:
Renaming a directory might fail with an EPERM error on Windows if the
target already exists. Therefore we stop creating the target directory
before the rename (this was probably the main problem) and also ensure
that it does not exist for whatever other reasons.

Related: https://github.com/SAP/ui5-project/pull/389
Resolves: https://github.com/SAP/ui5-tooling/issues/487

### DIFF
--- a/lib/ui5Framework/npm/Installer.js
+++ b/lib/ui5Framework/npm/Installer.js
@@ -79,9 +79,22 @@ class Installer {
 						await rimraf(stagingDir);
 					}
 
+					// Check whether target dir already exists and remove it.
+					// A target directory already existing but missing a package.json should
+					// never happen. However, we want to be *really* sure that there is no target
+					// directory so that the rename operation won't have any no trouble.
+					if (await this._pathExists(targetDir)) {
+						const rimraf = promisify(require("rimraf"));
+						log.verbose(`Removing existing targed directory at ${targetDir}...`);
+						await rimraf(targetDir);
+					}
+
 					log.verbose(`Installing ${pkgName} in version ${version} to ${stagingDir}...`);
 					await this._registry.extractPackage(pkgName, version, stagingDir);
-					await mkdirp(targetDir);
+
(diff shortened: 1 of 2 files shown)


RandomByte:
@matz3 kindly retest in your Windows reproduction setup

matz3:
Validated by manually replacing the file within node_modules.

lib/ui5Framework/npm/Installer.js
```diff
@@ -79,9 +79,22 @@ class Installer {
 						await rimraf(stagingDir);
 					}
 
+					// Check whether target dir already exists and remove it.
+					// A target directory already existing but missing a package.json should
+					// never happen. However, we want to be *really* sure that there is no target
+					// directory so that the rename operation won't have any no trouble.
+					if (await this._pathExists(targetDir)) {
+						const rimraf = promisify(require("rimraf"));
+						log.verbose(`Removing existing targed directory at ${targetDir}...`);
```
matz3, line 88:
```suggestion
						log.verbose(`Removing existing target directory at ${targetDir}...`);
```

lib/ui5Framework/npm/Installer.js
```diff
@@ -79,9 +79,22 @@ class Installer {
 						await rimraf(stagingDir);
 					}
 
+					// Check whether target dir already exists and remove it.
+					// A target directory already existing but missing a package.json should
+					// never happen. However, we want to be *really* sure that there is no target
+					// directory so that the rename operation won't have any no trouble.
```
matz3, line 85: Just as a note: This is especially important to fix the broken setup with CLI v2.9.0.

matz3 (changes requested):

lib/ui5Framework/npm/Installer.js
RandomByte: Right 👍

matz3 (approved):

PR #450, opened 2022-04-01
[FIX] lib/normalizer: Allow absolute windows paths and multiple parameters

flovogt:
This PR fixes the usages of absolute windows file paths e.g. `static:C:/myfolderA/myfileA` and offers a new way of passing multiple parameters to the translators e.g. `static:myfolderA/myfileA,myfolderB/myfileB`

### DIFF
--- a/lib/normalizer.js
+++ b/lib/normalizer.js
@@ -60,15 +60,14 @@ const Normalizer = {
 		log.verbose("Building dependency tree...");
 
 		let translatorParams = [];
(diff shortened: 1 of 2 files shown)


flovogt:
> 

As far as I remember we figured out that it was never working for windows machines, so we fixed it. And we had an idea to use the static translator property for an additional value but then we used another option. So we can close it. Fine for me.

PR #463, opened 2022-07-25
[FIX] Fix build manifest creation

RandomByte:


test/lib/build/ProjectBuilder.js
```diff
@@ -471,6 +471,214 @@ test.serial("_buildProject: Requested project is not added to readers", async (t
...
+	t.is(writerMock.write.getCall(0).args[0], "build manifest resource", "Write got called with build manifest");
+	t.is(writerMock.write.getCall(1).args[0], dummyResources[1], "Write got called with second resource");
+	t.is(writerMock.write.getCall(2).args[0], dummyResources[2], "Write got called with third resource");
+});
+
+test("_writeResults: Do not create build manifest for non-root project", async (t) => {
+	t.context.getRootTypeStub = sinon.stub().returns("library");
+	const {graph} = t.context;
+
+	const ProjectBuilder = mock.reRequire("../../../lib/build/ProjectBuilder");
```
matz3, line 636: Is this re-require necessary? If so, I'd rather expect this to be a `serial` test

matz3 (dismissed):

test/lib/build/ProjectBuilder.js
RandomByte: You're right, thanks! 👍

matz3 (approved):

PR #467, opened 2022-07-28
[INTERNAL] TaskRunner: Use new 'perf' log level instead of env variable

RandomByte:


flovogt:
Do we not have to wait for the next logger release and then increase the logger dependency here before merging?

RandomByte:
Already released: https://github.com/SAP/ui5-logger/releases/tag/v3.0.1-alpha.2

flovogt (approved):
LGTM

PR #468, opened 2022-08-04
[FIX] Specification: Fix migration for legacy projects that are not applications or libraries

RandomByte:
During the migration of legacy projects (defining specVersion < 2), the
propertiesFileSourceEncoding configuration was always added, even though
it is only relevant for applications and libraries. This lead to
validation errors for legacy projects of type "module" or extensions.

### DIFF
--- a/lib/specifications/types/extensions/ServerMiddleware.js
+++ b/lib/specifications/types/extensions/ServerMiddleware.js
@@ -23,6 +23,7 @@ class ServerMiddleware extends Extension {
 		if (/--\d+$/.test(this.getName())) {
 			throw new Error(`Server middleware name must not end with '--<number>'`);
 		}
+		// TODO: Check that paths exist
 	}
 }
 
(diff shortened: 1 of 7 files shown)


lib/specifications/types/extensions/ServerMiddleware.js
```diff
@@ -23,6 +23,7 @@ class ServerMiddleware extends Extension {
 		if (/--\d+$/.test(this.getName())) {
 			throw new Error(`Server middleware name must not end with '--<number>'`);
 		}
+		// TODO: Check that paths exist
```
flovogt, line 26: Is this a todo for this PR?

lib/specifications/types/extensions/Task.js
```diff
@@ -23,6 +23,7 @@ class Task extends Extension {
 		if (/--\d+$/.test(this.getName())) {
 			throw new Error(`Task name must not end with '--<number>'`);
 		}
+		// TODO: Check that paths exist
```
flovogt, line 26: Is this a todo for this PR?

flovogt (approved):

lib/specifications/types/extensions/ServerMiddleware.js
RandomByte: No, those are for another time 🙂

PR #470, opened 2022-08-18
[INTERNAL] use eslint-plugin-ava and cleanup test-suite

flovogt:
Introducing [eslint-plugin-ava](https://github.com/avajs/eslint-plugin-ava/) to avoid common pitfalls when writing powerful ava tests.

- Add plugin to eslint config and package.json
- Restructure current code base to latest recommendations

### DIFF
--- a/test/lib/build/ProjectBuilder.js
+++ b/test/lib/build/ProjectBuilder.js
@@ -1,9 +1,7 @@
 const test = require("ava");
-const sinon = require("sinon");
+const sinonGlobal = require("sinon");
 const path = require("path");
 const mock = require("mock-require");
-
-const ProjectBuilder = require("../../../lib/build/ProjectBuilder");
 
 function noop() {}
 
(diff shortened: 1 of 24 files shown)


test/lib/build/ProjectBuilder.js
```diff
@@ -75,7 +75,7 @@ test.afterEach.always(() => {
 	mock.stopAll();
 });
 
-test("Missing graph parameters", async (t) => {
+test.serial("Missing graph parameters", (t) => {
```
matz3, line 78: Is there a reason why `serial` is added for some tests?

test/lib/build/ProjectBuilder.js
```diff
@@ -700,7 +700,7 @@ test("_executeCleanupTasks", async (t) => {
 	const builder = new ProjectBuilder(graph, taskRepository);
 
 	const executeCleanupTasksStub = sinon.stub(builder._buildContext, "executeCleanupTasks");
-	builder._executeCleanupTasks();
+	await builder._executeCleanupTasks();
```
matz3, line 703: good catch

test/lib/build/ProjectBuilder.js
flovogt: Currently, other tests are failing when I remove the `async` identifier from the test function. The `serial` functions fixes this order issue. But I was not able to identify why the other test has any connection to these tests

matz3: The problem seems to be in the `afterEach` handler. It uses the global sinon sandbox to restore stubs, which affects also other tests. So either all tests should use serial or (better) they should use a local sinon sandbox that is the reset in the afterEach. I think we used this already once somewhere (instead of making all tests serial which is more common in our tests as it needs to be done anyways when shared modules are mocked/stubbed).

flovogt: Refactored by using the sandbox concept.

PR #514, opened 2022-11-17
[FEATURE] TaskUtil: Enhance interface provided to custom tasks defining specVersion 3.0 and later

RandomByte:
New API is only available to custom tasks defining specVerson 3.0 and
later.

### DIFF
--- a/lib/build/TaskRunner.js
+++ b/lib/build/TaskRunner.js
@@ -1,4 +1,5 @@
 import composeTaskList from "./helpers/composeTaskList.js";
+import logger from "@ui5/logger";
 
 /**
  * TaskRunner
@@ -279,11 +280,18 @@ class TaskRunner {
 						{@ui5/fs/DuplexCollection} parameters.workspace DuplexCollection to read and write files
 						{@ui5/fs/AbstractReader} parameters.dependencies
 							Reader or Collection to read dependency files
-						{Object} parameters.taskUtil Specification Version dependent interface to a
-							[TaskUtil]{@link @ui5/project/build/helpers/TaskUtil} instance
+						{@ui5/project/build/helpers/TaskUtil} parameters.taskUtil Specification Version dependent
+							interface of a [TaskUtil]{@link @ui5/project/build/helpers/TaskUtil} instance
+						{@ui5/logger/GroupLogger} [parameters.log] Logger instance to use by the custom task.
+							This method is only available to custom task extensions defining
+	 						<b>Specification Version 3.0 and above</b>.
 						{Object} parameters.options Options
 						{string} parameters.options.projectName Project name
-						{string} [parameters.options.projectNamespace] Project namespace if available
+						{string|null} parameters.options.projectNamespace Project namespace if available
+						{string} [parameters.options.taskName] Runtime name of the task.
+							If a task is executed multiple times, a suffix is added to distinguish the executions.
+							This attribute is only available to custom task extensions defining
+	 						<b>Specification Version 3.0 and above</b>.
 						{string} [parameters.options.configuration] Task configuration if given in ui5.yaml
 					Returns:
 						{Promise<undefined>} Promise resolving with undefined once data has been written
@@ -302,8 +310,8 @@ class TaskRunner {
 				}
 
 				if (task.getSpecVersion() === "3.0") {
-					params.log = log;
-					params.taskName = newTaskName;
+					params.options.taskName = newTaskName;
+					params.log = logger.getGroupLogger(`builder:custom-task:${newTaskName}`);
 				}
 
 				const taskUtilInterface = taskUtil.getInterface(task.getSpecVersion());
--- a/lib/build/TaskRunner.js
+++ b/lib/build/TaskRunner.js
@@ -280,7 +280,7 @@ class TaskRunner {
 						{@ui5/fs/DuplexCollection} parameters.workspace DuplexCollection to read and write files
 						{@ui5/fs/AbstractReader} parameters.dependencies
 							Reader or Collection to read dependency files
-						{@ui5/project/build/helpers/TaskUtil} parameters.taskUtil Specification Version dependent
+						{@ui5/project/build/helpers/TaskUtil} parameters.taskUtil Specification Version-dependent
 							interface of a [TaskUtil]{@link @ui5/project/build/helpers/TaskUtil} instance
 						{@ui5/logger/GroupLogger} [parameters.log] Logger instance to use by the custom task.
 							This method is only available to custom task extensions defining
--- a/lib/specifications/ComponentProject.js
+++ b/lib/specifications/ComponentProject.js
@@ -26,8 +26,11 @@ class ComponentProject extends Project {
 
 	/* === Attributes === */
 	/**
-	* @public
-	*/
+	 * Get the project namespace
+	 *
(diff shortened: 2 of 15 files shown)


matz3 (approved):

lib/specifications/ComponentProject.js
```diff
@@ -26,8 +26,11 @@ class ComponentProject extends Project {
 
 	/* === Attributes === */
 	/**
-	* @public
-	*/
+	 * Get the project namespace
+	 *
+	 * @public
+	 * @returns {string} Project namespace in slash-notation (e.g. <code>my/project/name</code>)
```
KlattG, line 32:
```suggestion
	 * @returns {string} Project namespace in slash notation (e.g. <code>my/project/name</code>)
```

lib/specifications/Project.js
```diff
@@ -42,85 +42,176 @@ class Project extends Specification {
 
 	/* === Attributes === */
 	/**
-	* @public
-	*/
+	 * Get the project namespace. Returns null for projects that have none or multiple namespaces.
```
KlattG, line 45:
```suggestion
	 * Get the project namespace. Returns `null` for projects that have none or multiple namespaces,
```

lib/specifications/Project.js
```diff
@@ -42,85 +42,176 @@ class Project extends Specification {
...
+	 * For example Modules or Theme Libraries.
+	 *
+	 * @public
+	 * @returns {string|null} Project namespace in slash-notation (e.g. <code>my/project/name</code>) or null
```
KlattG, line 49:
```suggestion
	 * @returns {string|null} Project namespace in slash notation (e.g. <code>my/project/name</code>) or null
```

lib/specifications/Project.js
```diff
@@ -42,85 +42,176 @@ class Project extends Specification {
 
 	/* === Attributes === */
 	/**
-	* @public
-	*/
+	 * Get the project namespace. Returns null for projects that have none or multiple namespaces.
+	 * For example Modules or Theme Libraries.
```
KlattG, line 46:
```suggestion
	 * for example Modules or Theme Libraries.
```

lib/specifications/Specification.js
```diff
@@ -110,52 +110,69 @@ class Specification {
...
+	 * @returns {string} Project name
+	 */
 	getName() {
 		return this._name;
 	}
 
 	/**
-	* @public
-	*/
+	 * Get the project kind
```
KlattG, line 123: How does this differ from the 'project type' below?

lib/specifications/Specification.js
```diff
@@ -110,52 +110,69 @@ class Specification {
...
+	 *
+	 * @public
+	 * @returns {string} Project kind
```
KlattG, line 126: It's confusing to have both a 'kind' and 'type' of project. Can this be remedied?

lib/specifications/Specification.js
```diff
@@ -110,52 +110,69 @@ class Specification {
...
+	 * @returns {string} Project type
+	 */
 	getType() {
 		return this._type;
 	}
 
 	/**
-	* @public
-	*/
+	 * Get the project Specification Version
```
KlattG, line 143:
```suggestion
	 * Get the project's Specification Version
```

KlattG (changes requested):

lib/specifications/Specification.js
RandomByte: Thank you for pointing this out. I did a mistake by referring to a specification as a "project".

In fact, a specification can be either a "project", **or** an "extension" (this is defined by the *kind* property). There might be other kinds in the future too...

I updated the documentation in the latest commit: [`88375b7` (#514)](https://github.com/SAP/ui5-project/pull/514/commits/88375b78c2bcecb9192903a98febe1c4084609a4)

RandomByte: I agree it is confusing. But sadly we introduced the additional "kind" term over two years ago when we added support for extensions. So I think it is to late to change any of the two terms now.

KlattG: That's really unfortunate. I think it would make sense to at least explain these terms better. Maybe we could put a more detailed example in, together with the definitions - what do you think?

matz3 (approved):

lib/specifications/Specification.js
```diff
@@ -110,47 +110,49 @@ class Specification {
...
-	 * @returns {string} Project name
+	 * @returns {string} Specification name
 	 */
 	getName() {
 		return this._name;
 	}
 
 	/**
-	 * Get the project kind
+	 * Get the kind of this specification. For example <code>project</code> or <code>extension</code>
```
KlattG, line 123:
```suggestion
	 * Get the kind of this specification, for example <code>project</code> or <code>extension</code>
```

lib/specifications/Specification.js
```diff
@@ -110,47 +110,49 @@ class Specification {
...
+	 * @returns {string} Specification kind
 	 */
 	getKind() {
 		return this._kind;
 	}
 
 	/**
-	 * Get the project type
+	 * Get the type of this specification.
+	 * For example <code>application</code> or <code>library</code> in case of projects,
```
KlattG, line 134:
```suggestion
	 * Get the type of this specification,
	 * for example <code>application</code> or <code>library</code> in case of projects,
```

lib/specifications/Specification.js
```diff
@@ -110,47 +110,49 @@ class Specification {
...
-	 * @returns {string} Project Specification Version
+	 * @returns {string} Specification Version
 	 */
 	getSpecVersion() {
 		return this._specVersion;
 	}
 
 	/**
-	 * Get the project version as typically defined in the project's <code>package.json</code>
+	 * Get the specification's version, as typically defined in a <code>package.json</code>
```
KlattG, line 155: Specification Version ?

KlattG (changes requested):

lib/build/TaskRunner.js
```diff
@@ -279,11 +280,18 @@ class TaskRunner {
 						{@ui5/fs/DuplexCollection} parameters.workspace DuplexCollection to read and write files
 						{@ui5/fs/AbstractReader} parameters.dependencies
 							Reader or Collection to read dependency files
-						{Object} parameters.taskUtil Specification Version dependent interface to a
-							[TaskUtil]{@link @ui5/project/build/helpers/TaskUtil} instance
+						{@ui5/project/build/helpers/TaskUtil} parameters.taskUtil Specification Version-dependent
+							interface of a [TaskUtil]{@link @ui5/project/build/helpers/TaskUtil} instance
+						{@ui5/logger/GroupLogger} [parameters.log] Logger instance to use by the custom task.
```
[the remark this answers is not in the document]
RandomByte: Yes!

lib/specifications/Specification.js
RandomByte: Not in this case. This is the generic version of the entity that makes up this "Specification instance".

KlattG: "Get the specification's generic version, as typically defined in a <code>package.json</code>"

RandomByte: Done 👍

matz3 (approved):

PR #518, opened 2022-12-02
[INTERNAL] Use native fs.mkdir instead of mkdirp

matz3:
The recursive option is available since Node v10.12, which makes the
use of thirdparty packages obsolete for most use cases.

### DIFF
--- a/lib/ui5Framework/npm/Installer.js
+++ b/lib/ui5Framework/npm/Installer.js
@@ -1,5 +1,4 @@
 import path from "node:path";
-import mkdirp from "mkdirp";
 import fs from "graceful-fs";
 import {promisify} from "node:util";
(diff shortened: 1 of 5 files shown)


codeworrior (approved):
LGTM, although it is not really 'native', it's from the graceful-fs, isn't it?
Well, but the signature is the one of the native fs.mkdir, so I'm fine with the subject. 

BTW: does graceful-fs meanwhile also support the promisified version of fs? Oh - no, [it doesn't](https://github.com/isaacs/node-graceful-fs/issues/160), too bad.

matz3:
Yes, it's not directly the native fs, but graceful-fs just wraps it and doesn't add/change the recursive feature. So, under the hood it's still the native feature.

PR #520, opened 2022-12-03
[INTERNAL] Refactor requires build/requires dependencies determination

RandomByte:
* Move requiresBuild() logic from TaskRunner to ProjectBuildContext
    * This likely needs to be moved into a module of its own at some
      point. For now, TaskRunner was the wrong place. If a project does
      not need to be built, there is no need to instantiate a TaskRunner
      in the first place. This is now implemented.
* TaskRunner can now compile a list of all direct project dependencies
  required by the set of tasks that it will execute for the given build
  options.
* Custom tasks defining specVersion >= 3.0 can now provide a
  'determineRequiredDependencies' callback to define which dependencies
  they require access to.

JIRA: CPOUI5FOUNDATION-568

### DIFF
--- a/lib/build/TaskRunner.js
+++ b/lib/build/TaskRunner.js
@@ -131,7 +131,7 @@ class TaskRunner {
 	 * First compiles a list of all tasks that will be executed, then a list of all direct project
 	 * dependencies that those tasks require access to.
 	 *
-	 * @returns {Set} Returns a set containing the names of all required direct project dependencies
+	 * @returns {Set<string>} Returns a set containing the names of all required direct project dependencies
 	 */
 	async getRequiredDependencies() {
 		await this._initTasks();
@@ -327,7 +327,7 @@ class TaskRunner {
 				taskConfiguration: taskDef.configuration,
 				getDependenciesReader: () => {
 					// Create the dependencies reader on-demand
-					return this._createDependencyReader(requiredDependencies);
+					return this._createDependenciesReader(requiredDependencies);
 				},
 			}),
 			requiredDependencies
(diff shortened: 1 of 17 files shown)


lib/build/TaskRunner.js
```diff
@@ -120,17 +122,19 @@ class TaskRunner {
...
 		}
 	}
 
 	/**
-	 * Takes a list of tasks which should be executed from the available task list of the current builder
+	 * First compiles a list of all tasks that will be executed, then a list of all direct project
+	 * dependencies that those tasks require access to.
 	 *
-	 * @returns {Promise} Returns promise chain with tasks
+	 * @returns {Set} Returns a set containing the names of all required direct project dependencies
```
matz3, line 134:
```suggestion
	 * @returns {Set<string>} Returns a set containing the names of all required direct project dependencies
```

matz3 (commented):
Nice 👍🏻

test/lib/build/TaskRunner.js
```diff
@@ -2,6 +2,7 @@ import test from "ava";
 import sinonGlobal from "sinon";
 import esmock from "esmock";
 import logger from "@ui5/logger";
+logger.setLevel("perf");
```
flovogt, line 5: why do we have to set the log level here?

lib/build/TaskRunner.js
```diff
@@ -214,138 +200,211 @@ class TaskRunner {
...
+		this._tasks[newTaskName] = {
+			task: this._createCustomTaskWrapper({
+				task,
+				project,
+				taskUtil,
+				taskName: newTaskName,
+				taskConfiguration: taskDef.configuration,
+				getDependenciesReader: () => {
+					// Create the dependencies reader on-demand
+					return this._createDependencyReader(requiredDependencies);
```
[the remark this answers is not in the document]
RandomByte: Done

test/lib/build/TaskRunner.js
RandomByte: In function `_executeTask`, we have some code that is only executed when log-level "perf" is activated

flovogt (approved):

matz3 (approved):

PR #523, opened 2022-12-07
[INTERNAL] Use new taskRepository API

RandomByte:
* Move requiresBuild() from TaskRunner to ProjectBuildContext
    * This likely needs to be moved into a module of its own at some
      point. For now, TaskRunner was the wrong place. If a project does
      not need to be built, there is no need to instantiate a TaskRunner
      in the first place. This is now implemented.
* TaskRunner can now compile a list of all direct project dependencies
  required by the set of tasks that it will execute for the given build
  options.
* Custom tasks defining specVersion >= 3.0 can now provide a
  'determineRequiredDependencies' callback to define which dependencies
  they require access to.

### DIFF
--- a/lib/build/ProjectBuilder.js
+++ b/lib/build/ProjectBuilder.js
@@ -34,11 +34,12 @@ class ProjectBuilder {
 	 * Executes a project build, including all necessary or requested dependencies
 	 *
 	 * @public
-	 * @param {@ui5/project/graph/ProjectGraph} graph Project graph
-	 * @param {@ui5/builder/tasks/taskRepository} taskRepository Task Repository module
-	 * @param {@ui5/project/build/ProjectBuilder~BuildConfiguration} [buildConfig] Build configuration
+	 * @param {object} parameters
+	 * @param {@ui5/project/graph/ProjectGraph} parameters.graph Project graph
+	 * @param {@ui5/project/build/ProjectBuilder~BuildConfiguration} [parameters.buildConfig] Build configuration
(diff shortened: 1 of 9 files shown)


matz3 (dismissed):

lib/build/ProjectBuilder.js
```diff
@@ -34,11 +35,12 @@ class ProjectBuilder {
...
-	 * @param {@ui5/project/graph/ProjectGraph} graph Project graph
-	 * @param {@ui5/builder/tasks/taskRepository} taskRepository Task Repository module
-	 * @param {@ui5/project/build/ProjectBuilder~BuildConfiguration} [buildConfig] Build configuration
+	 * @param {object} parameters
+	 * @param {@ui5/project/graph/ProjectGraph} parameters.graph Project graph
+	 * @param {@ui5/project/build/ProjectBuilder~BuildConfiguration} [parameters.buildConfig] Build configuration
+	 * @param {@ui5/builder/tasks/taskRepository} parameters.taskRepository Task Repository module to use
 	 */
-	constructor(graph, taskRepository, buildConfig) {
+	constructor({graph, buildConfig, taskRepository}) {
```
flovogt, line 43: changing the constructor here leads to a breaking change. So commit message should reflect his.

flovogt (approved):

RandomByte:
Rebased onto latest main

flovogt (approved):
LGTM. Awesome 🤩

PR #535, opened 2022-12-23
[INTERNAL] Specification#getRootReader: Use .gitignore excludes by default

RandomByte:
Depends on https://github.com/SAP/ui5-fs/pull/463

### DIFF
--- a/lib/specifications/Specification.js
+++ b/lib/specifications/Specification.js
@@ -216,17 +216,21 @@ class Specification {
 
 	/* === Resource Access === */
 	/**
-	* Get a [ReaderCollection]{@link @ui5/fs/ReaderCollection} for the root directory of the specification.
-	* Resource readers always use POSIX-style
-	*
-	* @public
-	* @returns {@ui5/fs/ReaderCollection} Reader collection
+	 * Get a [ReaderCollection]{@link @ui5/fs/ReaderCollection} for the root directory of the specification.
+	 * Resource readers always use POSIX-style
+	 *
+	 * @public
+	 * @param {object} [parameters] Parameters
+	 * @param {object} [parameters.useGitignore=false]
+	 *   Whether to apply any excludes defined in an optional .gitignore in the root directory
+	 * @returns {@ui5/fs/ReaderCollection} Reader collection
 	*/
-	getRootReader() {
+	getRootReader({useGitignore=true} = {}) {
 		return createReader({
 			fsBasePath: this.getPath(),
 			virBasePath: "/",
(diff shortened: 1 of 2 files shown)


lib/specifications/Specification.js
```diff
@@ -216,17 +216,21 @@ class Specification {
...
-	* Resource readers always use POSIX-style
-	*
-	* @public
-	* @returns {@ui5/fs/ReaderCollection} Reader collection
+	 * Get a [ReaderCollection]{@link @ui5/fs/ReaderCollection} for the root directory of the specification.
+	 * Resource readers always use POSIX-style
+	 *
+	 * @public
+	 * @param {object} [parameters] Parameters
+	 * @param {object} [parameters.useGitignore=false]
```
flovogt, line 224: its `true` not `false`

test/lib/specifications/Specification.js
```diff
@@ -37,6 +38,14 @@ test.afterEach.always((t) => {
 	sinon.restore();
 });
 
+test("Specification can't be instantiated", (t) => {
+	t.throws(() => {
+		new Specification();
+	}, {
```
flovogt, line 44: 🎉

test/lib/specifications/Specification.js
```diff
@@ -299,3 +308,88 @@ test("Invalid specVersion", async (t) => {
...
+
+test("getRootReader: Custom parameters", async (t) => {
+	const createReaderStub = sinon.stub();
+	const Specification = await esmock("../../../lib/specifications/Specification.js", {
+		"@ui5/fs/resourceFactory": {
+			createReader: createReaderStub
+		}
+	});
+
+	class DummySpecification extends Specification {
```
flovogt, line 359: dummy specification could be defined only once on top-level in the test file. so its' not necessary to define it twice.

flovogt (changes requested):

test/lib/specifications/Specification.js
RandomByte: Not in this case, since it's extending a mocked `Specification` reference, with a `createReader` stub.

lib/specifications/Specification.js
RandomByte: Done 👍

test/lib/specifications/Specification.js
flovogt: true...missed that one. but could be extracted to a separate function where the test context is given and this function can be called from the test need this dummyspec

RandomByte: Done 👌

flovogt (approved):

PR #536, opened 2022-12-27
[INTERNAL] TaskRunner: Custom Tasks receive 'dependencies' reader in most cases

RandomByte:
Always provide a 'dependencies'-reader to custom tasks, unless they
define Specification Version >=3.0 and do not provide a
determineRequiredDependencies callback.

The latter being an indicator that either dependencies are really not
required by the task or that the concept of the callback is not yet
known or applied incorrectly. By not providing any dependencies reader
in this case, the task will face an exception if access to the reader is
still attempted, for example via 'dependencies.byGlob()' (=> 'byGlob of
undefined').

At the same time, tasks implementing the callback but not requesting any
dependencies will receive an empty 'dependencies'-reader. This can ease
the implementation efforts in the task.

Resolves https://github.com/SAP/ui5-tooling/issues/732

### DIFF
--- a/test/lib/build/TaskRunner.js
+++ b/test/lib/build/TaskRunner.js
@@ -851,6 +851,67 @@ test("Custom task with specVersion 3.0", async (t) => {
 	}, "Task got called with one argument");
 });
 
+test("Custom task with specVersion 3.0 and no requiredDependenciesCallback", async (t) => {
+	const {sinon, graph, taskUtil, taskRepository, TaskRunner} = t.context;
+	const taskStub = sinon.stub();
+	const specVersionGteStub = sinon.stub().returns(true);
+	const mockSpecVersion = {
+		toString: () => "3.0",
+		gte: specVersionGteStub
+	};
+
+	const getRequiredDependenciesCallbackStub = sinon.stub().resolves(undefined);
+
+	graph.getExtension.returns({
+		getName: () => "custom task name",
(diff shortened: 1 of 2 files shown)


flovogt (approved):

test/lib/build/TaskRunner.js
```diff
@@ -851,6 +851,67 @@ test("Custom task with specVersion 3.0", async (t) => {
 	}, "Task got called with one argument");
 });
 
+test("Custom task with specVersion 3.0 and no requiredDependenciesCallback", async (t) => {
+	const {sinon, graph, taskUtil, taskRepository, TaskRunner} = t.context;
+	const taskStub = sinon.stub();
+	const specVersionGteStub = sinon.stub().returns(true);
+	const mockSpecVersion = {
```
flovogt, line 858: do we have a test for specVersion < 3.0?

RandomByte: Yes. I was surprised I missed to add this test, all the other combinations seem to be covered.

specVersion < 3.0 without callback: https://github.com/SAP/ui5-project/pull/536/files/c34a6628f3c75f107ce65fd41f05aba663a998e7#diff-27d98157bc20195d53765868ae91b0d166a863e5c582861f2bfa316c7a0bcbd1R631

flovogt: Ack

PR #494, opened 2022-11-04
[FEATURE] Add 'UI5 Workspace' Support

RandomByte:
Implementing https://github.com/SAP/ui5-tooling/pull/157

PR #543, opened 2023-01-12
[INTERNAL] ui5-workspace.yaml schema

d3xter666:
JIRA: CPOUI5FOUNDATION-584

Provide ui5-workspace.yaml schema that complies with the [local dependency resolution spec](https://github.com/SAP/ui5-tooling/blob/rfc-local-dependency-resolution/rfcs/0006-local-dependency-resolution.md#workspace-configuration-file)

For further details, check the [RFC 0006 Local Dependency Resolution](https://github.com/SAP/ui5-tooling/blob/rfc-local-dependency-resolution/rfcs/0006-local-dependency-resolution.md)

### DIFF
--- a/lib/validation/ValidationError.js
+++ b/lib/validation/ValidationError.js
@@ -49,7 +49,14 @@ class ValidationError extends Error {
 			separator += chalk.grey.dim("\u2500".repeat(process.stdout.columns || 80));
 		}
 		separator += "\n\n";
-		let message = chalk.red(`Invalid configuration for project ${this.project.id}`) + "\n\n";
+		let message;
+
+		if (this.project) { // ui5-workspace.yaml is project independent, so in that case, no project is available
+			message = chalk.red(`Invalid ui5.yaml configuration for project ${this.project.id}`) + "\n\n";
+		} else {
+			message = chalk.red(`Invalid ui5-workspace.yaml configuration.`) + "\n\n";
+		}
+
 		message += this.errors.map((error) => {
 			return this.formatError(error);
 		}).join(separator);
--- a/lib/validation/ValidationError.js
+++ b/lib/validation/ValidationError.js
@@ -49,7 +49,7 @@ class ValidationError extends Error {
 			separator += chalk.grey.dim("\u2500".repeat(process.stdout.columns || 80));
 		}
 		separator += "\n\n";
-		let message = chalk.red(`Invalid ui5.yaml configuration for project ${this.project.id}`) + "\n\n";
+		let message = chalk.red(`Invalid configuration for project ${this.project.id}`) + "\n\n";
 		message += this.errors.map((error) => {
 			return this.formatError(error);
 		}).join(separator);
--- a/lib/validation/ValidationError.js
+++ b/lib/validation/ValidationError.js
@@ -54,7 +54,7 @@ class ValidationError extends Error {
 		if (this.project) { // ui5-workspace.yaml is project independent, so in that case, no project is available
 			message = chalk.red(`Invalid ui5.yaml configuration for project ${this.project.id}`) + "\n\n";
 		} else {
-			message = chalk.red(`Invalid ui5-workspace.yaml configuration.`) + "\n\n";
+			message = chalk.red(`Invalid workspace configuration.`) + "\n\n";
 		}
 
 		message += this.errors.map((error) => {
--- a/lib/validation/validator.js
+++ b/lib/validation/validator.js
@@ -79,13 +79,16 @@ class Validator {
 const validator = Object.create(null);
 
 async function _validate(schemaName, options) {
-	if (!validator[schemaName]) {
-		validator[schemaName] = await (async () => {
-			const {default: Ajv} = await import("ajv");
-			const {default: ajvErrors} = await import("ajv-errors");
-			return new Validator({Ajv, ajvErrors, schemaName});
-		})();
-	}
+	validator[schemaName] = await (async () => {
+		if (validator[schemaName]) {
+			return validator[schemaName];
+		}
+
+		const {default: Ajv} = await import("ajv");
+		const {default: ajvErrors} = await import("ajv-errors");
+		return new Validator({Ajv, ajvErrors, schemaName});
+	})();
+
 	await validator[schemaName].validate(options);
 }
 
--- a/lib/validation/validator.js
+++ b/lib/validation/validator.js
@@ -1,8 +1,36 @@
 import {fileURLToPath} from "node:url";
 import {readFile} from "node:fs/promises";
 
+/**
+ * @module @ui5/project/validation/validator
+ * @description A collection of validation related APIs
+ * @public
+ */
+
+/**
+ * @enum {string}
+ * @private
+ * @readonly
+ */
+const SCHEMA_VARIANTS = {
+	"ui5": "ui5.json",
+	"ui5.json": "ui5.json",
+	"ui5-workspace": "ui5-workspace.json",
+	"ui5-workspace.json": "ui5-workspace.json",
+};
+
 class Validator {
-	constructor({Ajv, ajvErrors}) {
+	constructor({Ajv, ajvErrors, schemaName}) {
+		if (!schemaName && !SCHEMA_VARIANTS[schemaName]) {
+			throw new Error(
+				`"schemaName" is missing or incorrect. The available schemaName variants are ${Object.keys(
+					SCHEMA_VARIANTS
+				)}`
+			);
+		}
+
+		this._schemaName = SCHEMA_VARIANTS[schemaName];
+
 		this.ajv = new Ajv({
 			allErrors: true,
 			jsonPointers: true,
@@ -11,17 +39,17 @@ class Validator {
 		ajvErrors(this.ajv);
 	}
 
-	_compileSchema(schemaName) {
-		this._compiling = this._compiling || {};
+	_compileSchema() {
+		const schemaName = this._schemaName;
 
-		if (!this._compiling[schemaName]) {
-			this._compiling[schemaName] = Promise.resolve().then(async () => {
+		if (!this._compiling) {
+			this._compiling = Promise.resolve().then(async () => {
 				const schema = await Validator.loadSchema(schemaName);
 				const validate = await this.ajv.compileAsync(schema);
 				return validate;
 			});
 		}
-		return this._compiling[schemaName];
+		return this._compiling;
 	}
 
 	async validate({config, project, yaml, schemaName}) {
@@ -52,23 +80,6 @@ class Validator {
 
 const validator = {};
 
-/**
- * @module @ui5/project/validation/validator
- * @description A collection of validation related APIs
- * @public
- */
-
-/**
- * @enum {string}
- * @private
- * @readonly
- */
-const SCHEMA_VARIANTS = {
-	"ui5": "ui5.json",
-	"ui5.json": "ui5.json",
-	"ui5-workspace": "ui5-workspace.json",
-	"ui5-workspace.json": "ui5-workspace.json",
-};
 
 /**
  * Validates the given configuration.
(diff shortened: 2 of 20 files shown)


lib/validation/validator.js
```diff
@@ -1,37 +1,33 @@
...
 			allErrors: true,
 			jsonPointers: true,
-			loadSchema
+			loadSchema: Validator.loadSchema
 		});
 		ajvErrors(this.ajv);
 	}
 
-	_compileSchema() {
+	_compileSchema(schemaName) {
```
matz3, line 14: The current implementation of _compileSchema doesn't support multiple schemas as it caches the compiled fn in a promise. I've added a failing test to demonstrate this.

I think the easiest solution would be to just have two independent instances of `Validator`, one for each schema. This way it's ensure that they don't interfere with each other. My new test would then become obsolete.

matz3 (changes requested):

lib/validation/validator.js
d3xter666: How about now?
It's been limited to the schemas that we support, but still uses (almost) the same interface.
The caching issues that we had, now should be addressed and not face them anymore. Even if we try to extend the validator with some other schema, now should be easier- just extend the enum with the available schemas

matz3: Issues look resolved 👍🏻
However I now feel like, as the two approaches have been combined, it is a bit over-engineered.

When we use separate instances for each schema, why don't we pass the schema name to the constructor so that this instance always and only validates that specific schema?

On the other hand, when adjusting _compileSchema to work with multiple schemas I don't see a reason to have multiple instances. But I might have missed some detail here.

d3xter666: The shcemaName as an argument to the constructor looks like the correct approach. In that case, we'll avoid ambiguities about which schema is actually being validated. I'll refactor it. Thanks!

test/lib/validation/validator.js
```diff
@@ -39,44 +39,58 @@ test("validate function calls Validator#validate method", async (t) => {
 	t.deepEqual(validateStub.getCall(0).args, [{config, project, yaml, schemaName: "ui5.json"}]);
 });
 
-test("Validator#_compileSchema with two simultaneous calls for different schemas", async (t) => {
+test("Validator requires schemaName", (t) => {
```
matz3, line 42: Please also add a test with an invalid schemaName.

lib/validation/validator.js
```diff
@@ -1,8 +1,36 @@
...
+/**
+ * @enum {string}
+ * @private
+ * @readonly
+ */
+const SCHEMA_VARIANTS = {
+	"ui5": "ui5.json",
+	"ui5.json": "ui5.json",
+	"ui5-workspace": "ui5-workspace.json",
+	"ui5-workspace.json": "ui5-workspace.json",
```
matz3, line 19: Why are there two different keys per value?

lib/validation/validator.js
```diff
@@ -1,8 +1,36 @@
...
 
+/**
+ * @module @ui5/project/validation/validator
+ * @description A collection of validation related APIs
+ * @public
+ */
+
+/**
+ * @enum {string}
+ * @private
```
matz3, line 12: this is referenced from public API, so the enum must be public. Reference should be checked in JSDoc.

matz3 (changes requested):

lib/validation/validator.js
d3xter666: It seems like it's more convenient if the end user could be able to provide both variants- with the extension and without it. 
Of course, we might want to limit this to either of those

lib/validation/validator.js
d3xter666: So, we'd need to export it, too, right?

lib/validation/validator.js
```diff
@@ -70,18 +93,29 @@ let validator;
...
+	if (!SCHEMA_VARIANTS[schemaName]) {
+		throw new Error(
+			`"schemaName" is missing or incorrect. The available schemaName variants are ${Object.keys(
+				SCHEMA_VARIANTS
+			)}`
+		);
+	}
+
+	if (!validator[schemaName]) {
 		const {default: Ajv} = await import("ajv");
```
RandomByte, line 114: `validator` remains empty until the asynchronous imports are done. This could be a race condition if multiple validations with the same schema are started synchronously. The only real impact would be bad performance, but a fix should be easy enough.

A possible solution from a similar case can be found here: https://github.com/SAP/ui5-project/blob/99a440d0faa36d31b44e47d7866a2c4c0fd8e3f8/lib/graph/Module.js#L113-L123

lib/validation/validator.js
```diff
@@ -70,18 +93,29 @@ let validator;
...
  * @throws {@ui5/project/validation/ValidationError}
  *   Rejects with a {@link @ui5/project/validation/ValidationError ValidationError}
  *   when the validation fails.
  * @returns {Promise<undefined>} Returns a Promise that resolves when the validation succeeds
  */
 export async function validate(options) {
-	if (!validator) {
+	const schemaName = options.schemaName;
+
+	if (!SCHEMA_VARIANTS[schemaName]) {
```
RandomByte, line 105: This check is already done in the Validator constructor. Maybe we don't need both?

Since the other check is covered by tests while this one [isn't](https://coveralls.io/builds/56007649/source?filename=lib%2Fvalidation%2Fvalidator.js#L106), maybe we can remove it?

lib/validation/validator.js
```diff
@@ -1,28 +1,48 @@
...
+};
 
 class Validator {
-	constructor({Ajv, ajvErrors}) {
+	constructor({Ajv, ajvErrors, schemaName}) {
+		if (!schemaName && !SCHEMA_VARIANTS[schemaName]) {
+			throw new Error(
+				`"schemaName" is missing or incorrect. The available schemaName variants are ${Object.keys(
+					SCHEMA_VARIANTS
+				)}`
```
RandomByte, line 26:
```suggestion
				)}.join(", ")`
```

For better readability?

lib/validation/validator.js
```diff
@@ -46,15 +66,18 @@ class Validator {
...
+		const filePath = schemaPath.replace("http://ui5.sap/schema/", "");
+		const schemaFile = await readFile(
+			fileURLToPath(new URL(`./schema/${filePath}`, import.meta.url)), {encoding: "utf8"}
+		);
+		return JSON.parse(schemaFile);
+	}
 }
 
-let validator;
+const validator = {};
```
RandomByte, line 79: Either use a prototype-less object using `Object.create(null)` or a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

Matthias changed most of our lists to simple objects in https://github.com/SAP/ui5-project/commit/d4044ecd06dd6fc438a660d5d0af91a46b0e8c62 and related commits. Meanwhile, we try to use Set and Map wherever possible.

RandomByte (changes requested):
Just some minor things. I'll try and test this PR together with my local workspace demo setup.

lib/validation/ValidationError.js
```diff
@@ -49,7 +49,7 @@ class ValidationError extends Error {
 			separator += chalk.grey.dim("\u2500".repeat(process.stdout.columns || 80));
 		}
 		separator += "\n\n";
-		let message = chalk.red(`Invalid ui5.yaml configuration for project ${this.project.id}`) + "\n\n";
+		let message = chalk.red(`Invalid configuration for project ${this.project.id}`) + "\n\n";
```
RandomByte, line 52: I validated this PR with https://github.com/SAP/ui5-project/pull/494 and noticed that I can't really have a project at hand when validating a workspace file (see https://github.com/SAP/ui5-project/pull/494/commits/2182c2905e836e36ee54f007759109c20dc7bfe1#diff-416284583829e166b5b22422140544273617f7ecdeda4e9cceb4f7e698890241R283).

In fact, the error message is also slightly misleading. Since we typically only refer to `ui5.yaml`-configuration as "project configuration", I would maybe expect something like `Invalid workspace configuration` (without referencing any project).

Maybe we can add an additional parameter to have a different error message for invalid workspace configurations?

lib/validation/validator.js
d3xter666: Thanks!

lib/validation/validator.js
d3xter666: After the latest refactoring, this went again to @private. The reasoning is that we now have 2 separate public functions for validation and this enum is behind their facade.

lib/validation/validator.js
```diff
@@ -78,8 +78,19 @@ class Validator {
 
 const validator = Object.create(null);
 
+async function _validate(schemaName, options) {
+	if (!validator[schemaName]) {
+		validator[schemaName] = await (async () => {
```
RandomByte, line 83: I think this still has potential for a race condition. 

The map entry is only set once the promise has resolved. Which is asynchronous. A second validate call might see the map as empty and start creating a new Validator instance itself.

Instead we should store the promise in the map and await on that.

lib/validation/validator.js
```diff
@@ -92,23 +103,34 @@ const validator = Object.create(null);
...
+ * @public
+ * @function
+ * @static
+ * @param {object} options
+ * @param {object} options.config ui5-workspace Configuration to validate
+ * @param {object} [options.yaml] YAML information
+ * @param {string} options.yaml.path Path of the YAML file
+ * @param {string} options.yaml.source Content of the YAML file
+ * @param {number} [options.yaml.documentIndex=0] Document index in case the YAML file contains multiple documents
+ * @param {SCHEMA_VARIANTS} options.schemaName The name of the schema to validate
```
RandomByte: This parameter has been removed

lib/validation/ValidationError.js
```diff
@@ -49,7 +49,14 @@ class ValidationError extends Error {
 			separator += chalk.grey.dim("\u2500".repeat(process.stdout.columns || 80));
 		}
 		separator += "\n\n";
-		let message = chalk.red(`Invalid configuration for project ${this.project.id}`) + "\n\n";
+		let message;
+
+		if (this.project) { // ui5-workspace.yaml is project independent, so in that case, no project is available
+			message = chalk.red(`Invalid ui5.yaml configuration for project ${this.project.id}`) + "\n\n";
+		} else {
+			message = chalk.red(`Invalid ui5-workspace.yaml configuration.`) + "\n\n";
```
RandomByte, line 57: Let's just say `Invalid workspace configuration`.

The actual file name might be different.

RandomByte (changes requested):

lib/validation/validator.js
d3xter666: Thanks! Removed it

lib/validation/validator.js
d3xter666: Hmmmm, I got it. I'll do it with a promise. Thanks!

d3xter666: I think now it should be ok

RandomByte: Looks good 👌

RandomByte (approved):
Verified it once again with my change and everything works as expected. Great work 🎉

test/lib/validation/validator.js
```diff
@@ -20,3 +39,100 @@ test.serial("validate function calls Validator#validate method", async (t) => {
...
+	});
+});
+
+test("Validator requires a valid schemaName", (t) => {
+	const {sinon, Validator} = t.context;
+
+	const Ajv = sinon.stub();
+	const ajvErrors = sinon.stub();
+	const invalidContructor = () => {
+		new Validator({Ajv, ajvErrors, schemName: "invalid schema name"});
```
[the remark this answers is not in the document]
d3xter666: Thanks a lot, @matz3 !

Prepared a fix for that: https://github.com/SAP/ui5-project/pull/567

PR #549, opened 2023-01-18
[INTERNAL] specification/Project: Make getSourcePath method public

RandomByte:
The method will also be provided through TaskUtil and MiddlewareUtil.

This should help use cases like in the ui5-middleware-fe-mockserver:
https://github.com/SAP/open-ux-odata/blob/main/packages/ui5-middleware-fe-mockserver/src/index.ts#L10


d3xter666:
LGTM

flovogt (approved):

PR #557, opened 2023-01-27
[INTERNAL] Specification: Make method getPath() public

RandomByte:


flovogt (dismissed):

flovogt:
So, getPath will be public, but not in taskutils?

RandomByte:
> So, getPath will be public, but not in taskutils?

Good point!

flovogt (approved):

PR #561, opened 2023-01-30
[INTERNAL] Allow framework projects use of framework resolver

RandomByte:
If the current root project is a framework project (identified by the scope in the package.json name), allow processing of it's framework dependencies by the ui5Framework helper module. Other framework projects in the graph are still ignored.

Framework dependencies of non-framework projects are always processed. Framework dependencies of framework projects have always been ignored. With this change, framework dependencies of the root project are always processed, no matter whether it's a framework projects or not.

This allows framework developers to retrieve all required dependencies in a given version. The version is either defined in the ui5.yaml (not recommended) or via the --framework-version CLI parameter.

If no version is specified, all required framework dependencies must be part of the presented graph or an error is thrown.

### DIFF
--- a/lib/graph/helpers/ui5Framework.js
+++ b/lib/graph/helpers/ui5Framework.js
@@ -69,7 +69,7 @@ const utils = {
 		const ui5Dependencies = [];
 		const rootProject = projectGraph.getRoot();
 		await projectGraph.traverseBreadthFirst(async ({project}) => {
-			if (project.isFrameworkProject()) {
+			if (project !== rootProject && project.isFrameworkProject()) {
 				// Ignoring UI5 Framework libraries in dependencies
 				return;
 			}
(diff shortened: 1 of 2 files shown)


lib/graph/helpers/ui5Framework.js
```diff
@@ -69,7 +69,7 @@ const utils = {
 		const ui5Dependencies = [];
 		const rootProject = projectGraph.getRoot();
 		await projectGraph.traverseBreadthFirst(async ({project}) => {
-			if (project.isFrameworkProject()) {
```
flovogt, line 72: can `project` be `undefined` or `null`?

RandomByte: No, this is guaranteed by the `traverseBreadthFirst´ method

flovogt (approved):

PR #562, opened 2023-01-30
[INTERNAL] azure: Use latest os and add node 19

flovogt:


d3xter666 (commented):
Just one thing: If OS versions are `latest`, then this would mean we do not have a stable OS, but always the latest. I'm not sure whether we have faced issues based on the OS or rather rely entirely on Node

RandomByte (approved):

PR #478, opened 2022-10-27
[FEATURE][PoC] Enable snapshot consumption from Maven repository

RandomByte:
**Usage:** Specify framework `SAPUI5` and a version ending with SNAPSHOT, i.e. `1.108.0-SNAPSHOT`. Then execute ui5 command with environment variable `SNAPSHOT_ENDPOINT_URL`, for example:

```sh
UI5_MAVEN_SNAPSHOT_ENDPOINT=https://<hostname>/artifactory/build-snapshots/ ui5 tree --verbose
```

**Outlook:** If we go ahead with implementing this, the "Snapshot Endpoint URL" environment variable should be replaced with a central configuration and accompanying CLI parameters to override the central configuration.

TODO: How to decide whether to consume pre-built or sources of dependencies?

For more details, see internal concept document.

PR #570, opened 2023-02-07
[FEATURE] Enable snapshot consumption from Maven repository

d3xter666:
~~This is just a rebase with `main` of [Enable snapshot consumption from Maven repository #569](https://github.com/SAP/ui5-project/pull/478). The original change had conflicts with the [workspace implementation](https://github.com/SAP/ui5-project/pull/494).~~

Implementation of snapshot consumption from Maven repository. This is the final implementation and is the continuation work from the base change: https://github.com/SAP/ui5-project/pull/478

~~The change also implements the RFC for a [Configuration](https://github.com/SAP/ui5-tooling/blob/rfc-configuration/rfcs/0013-configuration.md)~~
The configuration part has been moved into a separate change: https://github.com/SAP/ui5-project/pull/575

JIRA: CPOUI5FOUNDATION-581

### DIFF
--- a/lib/graph/helpers/ui5Framework.js
+++ b/lib/graph/helpers/ui5Framework.js
@@ -299,7 +299,7 @@ export default {
 		let version = options.versionOverride || frameworkVersion;
 
 		if (rootProject.isFrameworkProject() && !version) {
-			// If the root project is a framework project, and the framework version is not a snapshot version,
+			// If the root project is a framework project and no framework version is defined,
 			// all framework dependencies must either be already part of the graph or part of the workspace.
 			// A mixed setup of framework deps within the graph AND from the workspace is currently not supported.
 
--- a/lib/graph/helpers/ui5Framework.js
+++ b/lib/graph/helpers/ui5Framework.js
@@ -298,7 +298,7 @@ export default {
 		// framework version in their respective ui5.yaml
 		let version = options.versionOverride || frameworkVersion;
 
-		if (rootProject.isFrameworkProject() && (!version || !version.endsWith("-SNAPSHOT"))) {
+		if (rootProject.isFrameworkProject() && !version) {
 			// If the root project is a framework project, and the framework version is not a snapshot version,
 			// all framework dependencies must either be already part of the graph or part of the workspace.
 			// A mixed setup of framework deps within the graph AND from the workspace is currently not supported.
@@ -329,7 +329,7 @@ export default {
 		}
 
 		let Resolver;
-		if (version.endsWith("-SNAPSHOT")) {
+		if (version && version.endsWith("-SNAPSHOT")) {
 			Resolver = (await import("../../ui5Framework/Sapui5MavenSnapshotResolver.js")).default;
 		} else if (frameworkName === "OpenUI5") {
 			Resolver = (await import("../../ui5Framework/Openui5Resolver.js")).default;
@@ -368,8 +368,8 @@ export default {
 		const resolver = new Resolver({
 			cwd: rootProject.getRootPath(),
 			version,
+			providedLibraryMetadata,
 			...options.resolverConfig,
-			providedLibraryMetadata
 		});
 
 		let startTime;
--- a/lib/ui5Framework/AbstractResolver.js
+++ b/lib/ui5Framework/AbstractResolver.js
@@ -15,7 +15,7 @@ const SEMVER_VERSION_REGEXP = /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:
 
 // Reduced Semantic Versioning pattern
 // Matches MAJOR.MINOR as a simple version range to be resolved to the latest patch
-const VERSION_RANGE_REGEXP = /^(0|[1-9]\d*)\.(0|[1-9]\d*)$/;
+const VERSION_RANGE_REGEXP = /^(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-SNAPSHOT)?$/;
 
 /**
  * Abstract Resolver
@@ -210,15 +210,28 @@ class AbstractResolver {
 
 	static async resolveVersion(version, {ui5HomeDir, cwd} = {}) {
 		let spec;
-		if (version === "latest") {
+		const isSnapshotVersion = version.toLowerCase().endsWith("-snapshot");
+		if (version === "latest" || version === "latest-snapshot") {
+			// Use a wildcard to resolve to the latest available version
 			spec = "*";
-		} else if (VERSION_RANGE_REGEXP.test(version) || SEMVER_VERSION_REGEXP.test(version)) {
+		} else if (SEMVER_VERSION_REGEXP.test(version)) {
+			// Fully qualified version, can be used directly
 			spec = version;
+		} else if (VERSION_RANGE_REGEXP.test(version)) {
+			if (isSnapshotVersion) {
+				// For snapshot version ranges we need to insert a stand-in "x" for the patch level
+				// in order to make the semver check work: "1.112-SNAPSHOT" becomes "1.112.x-SNAPSHOT"
+				spec = version.replace(/-SNAPSHOT$/, ".x-SNAPSHOT");
+			} else {
+				spec = version;
+			}
 		} else {
 			throw new Error(`Framework version specifier "${version}" is incorrect or not supported`);
 		}
 		const versions = await this.fetchAllVersions({ui5HomeDir, cwd});
-		const resolvedVersion = semver.maxSatisfying(versions, spec);
+		const resolvedVersion = semver.maxSatisfying(versions, spec, {
+			includePrerelease: isSnapshotVersion
+		});
 		if (!resolvedVersion) {
 			if (semver.valid(spec)) {
 				if (this.name === "Sapui5Resolver" && semver.lt(spec, "1.76.0")) {
--- a/lib/ui5Framework/AbstractResolver.js
+++ b/lib/ui5Framework/AbstractResolver.js
@@ -32,7 +32,7 @@ class AbstractResolver {
 	 * @param {*} options options
 	 * @param {string} [options.version] Framework version to use. When omitted, all libraries need to be available
 	 * via <code>providedLibraryMetadata</code> parameter. Otherwise an error is thrown.
-	 * @param {string} [options.sources=false] Whether to install framework libraries as sources or
+	 * @param {boolean} [options.sources=false] Whether to install framework libraries as sources or
 	 * 					pre-built (with build manifest)
 	 * @param {string} [options.cwd=process.cwd()] Current working directory
 	 * @param {string} [options.ui5HomeDir="~/.ui5"] UI5 home directory location. This will be used to store packages,
--- a/lib/ui5Framework/AbstractResolver.js
+++ b/lib/ui5Framework/AbstractResolver.js
@@ -34,7 +34,7 @@ class AbstractResolver {
 	 * via <code>providedLibraryMetadata</code> parameter. Otherwise an error is thrown.
 	 * @param {string} [options.sources=false] Whether to install framework libraries as sources or
 	 * 					pre-built (with build manifest)
-	 * @param {string} [options.cwd=process.cwd()] Working directory to resolve configurations like .npmrc
+	 * @param {string} [options.cwd=process.cwd()] Current working directory
 	 * @param {string} [options.ui5HomeDir="~/.ui5"] UI5 home directory location. This will be used to store packages,
 	 * metadata and configuration used by the resolvers. Relative to `process.cwd()`
 	 * @param {object.<string, @ui5/project/ui5Framework/AbstractResolver~LibraryMetadataEntry>} [options.providedLibraryMetadata]
--- a/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
+++ b/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
@@ -43,9 +43,6 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 	constructor(options) {
 		super(options);
 
-		const snapshotEndpointUrl =
-			options.snapshotEndpointUrl ||
-			process.env.UI5_MAVEN_SNAPSHOT_ENDPOINT;
 		const {
 			cacheMode,
 			artifactsDir,
@@ -54,20 +51,10 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 			stagingDir,
 		} = options;
 
-		if (!snapshotEndpointUrl) {
-			// If we resolve the settings.xml at this point, we'd need to always ask the end
-			// user for confirmation. In some cases where the resources are already cached,
-			// this is not necessary and we could skip it as a real request to the repositry won't
-			// be made.
-			this._snapshotEndpointUrlCb = Sapui5MavenSnapshotResolver._resolveSnapshotEndpointUrl;
-			// TODO: Save resolved snapshot URL when .ui5rc Configuration is implemented
-		} else {
-			this._snapshotEndpointUrlCb = () => Promise.resolve(snapshotEndpointUrl);
-		}
-
 		this._installer = new Installer({
 			ui5HomeDir: this._ui5HomeDir,
-			snapshotEndpointUrlCb: this._snapshotEndpointUrlCb,
+			snapshotEndpointUrlCb:
+				Sapui5MavenSnapshotResolver.createSnapshotEndpointUrlCallback(options.snapshotEndpointUrl),
 			cacheMode,
 			artifactsDir,
 			packagesDir,
@@ -142,19 +129,35 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 			}),
 		};
 	}
-	static async fetchAllVersions({ui5HomeDir, cwd} = {}) {
+
+	static async fetchAllVersions({ui5HomeDir, cwd, snapshotEndpointUrl} = {}) {
 		const installer = new Installer({
 			cwd: cwd ? path.resolve(cwd) : process.cwd(),
 			ui5HomeDir: path.resolve(
 				ui5HomeDir || path.join(os.homedir(), ".ui5")
 			),
-			snapshotEndpointUrlCb: this._snapshotEndpointUrlCb,
+			snapshotEndpointUrlCb: Sapui5MavenSnapshotResolver.createSnapshotEndpointUrlCallback(snapshotEndpointUrl),
 		});
 		return await installer.fetchPackageVersions({
 			groupId: DIST_GROUP_ID,
 			artifactId: DIST_ARTIFACT_ID,
 		});
 	}
+
+	static createSnapshotEndpointUrlCallback(snapshotEndpointUrl) {
+		snapshotEndpointUrl = snapshotEndpointUrl || process.env.UI5_MAVEN_SNAPSHOT_ENDPOINT;
+
+		if (!snapshotEndpointUrl) {
+			// If we resolve the settings.xml at this point, we'd need to always ask the end
+			// user for confirmation. In some cases where the resources are already cached,
+			// this is not necessary and we could skip it as a real request to the repository won't
+			// be made.
+			return Sapui5MavenSnapshotResolver._resolveSnapshotEndpointUrl;
+		} else {
+			return () => Promise.resolve(snapshotEndpointUrl);
+		}
+	}
+
 	/**
 	 * Tries to detect whether ~/.m2/settings.xml exist, and if so, whether
 	 * the snapshot.build URL is extracted from there
--- a/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
+++ b/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
@@ -35,7 +35,7 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 	 * 	"off" (do not cache) and "force" (use cache only - no requests)
 	 * @param {string} [options.artifactsDir] Where to install Maven artifacts
 	 * @param {string} [options.packagesDir] Where to install packages
-	 * @param {string} [options.metadataDir] Where to store the metadata for Maven artifacts 
+	 * @param {string} [options.metadataDir] Where to store the metadata for Maven artifacts
 	 * @param {string} [options.stagingDir] The staging directory for artifacts and packages
 	 */
 	constructor(options) {
@@ -110,7 +110,8 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 		const metadata = await this.getLibraryMetadata(libraryName);
 		if (!metadata.gav) {
 			throw new Error(
-				`Metadata is missing GAV information. This might indicate an unsupported SNAPSHOT version.`
+				"Metadata is missing GAV (group, artifact and version) " +
+					"information. This might indicate an unsupported SNAPSHOT version."
 			);
 		}
 		const gav = metadata.gav.split(":");
@@ -151,7 +152,8 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 		});
 	}
 	/**
-	 * Tries to detect whether ~/.m2/settings.xml exist, and if so, whether the snapshot.build URL is extracted from there
+	 * Tries to detect whether ~/.m2/settings.xml exist, and if so, whether
+	 * the snapshot.build URL is extracted from there
 	 *
 	 * @param {string} [settingsXML=~/.m2/settings.xml] Path to the settings.xml.
 	 * 				If not provided, the default location is used
--- a/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
+++ b/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
@@ -52,7 +52,6 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 			packagesDir,
 			metadataDir,
 			stagingDir,
-			sources,
 		} = options;
 
 		if (!snapshotEndpointUrl) {
--- a/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
+++ b/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
@@ -36,7 +36,7 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 	constructor(options) {
 		super(options);
 
-		let snapshotEndpointUrl =
+		const snapshotEndpointUrl =
 			options.snapshotEndpointUrl ||
 			process.env.UI5_MAVEN_SNAPSHOT_ENDPOINT;
 		const {
@@ -48,19 +48,28 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 			sources,
 		} = options;
 
+		let snapshotEndpointResolverP;
 		if (!snapshotEndpointUrl) {
-			snapshotEndpointUrl = Sapui5MavenSnapshotResolver._resolveSnapshotEndpointUrl();
+			snapshotEndpointResolverP = Sapui5MavenSnapshotResolver._resolveSnapshotEndpointUrl();
 			// TODO: Save resolved snapshot URL when .ui5rc Configuration is implemented
+		} else {
+			snapshotEndpointResolverP = Promise.resolve(snapshotEndpointUrl);
 		}
 
-		this._installer = new Installer({
-			ui5HomeDir: this._ui5HomeDir,
-			snapshotEndpointUrl,
-			cacheMode,
-			artifactsDir,
-			packagesDir,
-			metadataDir,
-			stagingDir,
+		this._installerP = new Promise((resolve) => {
+			snapshotEndpointResolverP.then((snapshotUrl) => {
+				resolve(
+					new Installer({
+						ui5HomeDir: this._ui5HomeDir,
+						snapshotEndpointUrl: snapshotUrl,
+						cacheMode,
+						artifactsDir,
+						packagesDir,
+						metadataDir,
+						stagingDir,
+					})
+				);
+			});
 		});
 		this._loadDistMetadata = null;
 		this._sources = !!sources;
@@ -72,17 +81,18 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 				log.verbose(
 					`Installing ${DIST_ARTIFACT_ID} in version ${version}...`
 				);
-				const {pkgPath: distPkgPath} =
-					await this._installer.installPackage({
-						pkgName: DIST_PKG_NAME,
-						groupId: DIST_GROUP_ID,
-						artifactId: DIST_ARTIFACT_ID,
-						version,
-						classifier: "npm-sources",
-						extension: "zip",
-					});
 
-				return await this._installer.readJson(
+				const installer = await this._installerP;
+				const {pkgPath: distPkgPath} = await installer.installPackage({
+					pkgName: DIST_PKG_NAME,
+					groupId: DIST_GROUP_ID,
+					artifactId: DIST_ARTIFACT_ID,
+					version,
+					classifier: "npm-sources",
+					extension: "zip",
+				});
+
+				return await installer.readJson(
 					path.join(distPkgPath, "metadata.json")
 				);
 			});
@@ -111,6 +121,8 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 		if (!this._sources) {
 			pkgName += "-prebuilt";
 		}
+
+		const installer = await this._installerP;
 		return {
 			metadata: Promise.resolve({
 				id: pkgName,
@@ -119,7 +131,7 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 				optionalDependencies: metadata.optionalDependencies,
 			}),
 			// Trigger installation of package
-			install: this._installer.installPackage({
+			install: installer.installPackage({
 				pkgName,
 				groupId: gav[0],
 				artifactId: gav[1],
--- a/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
+++ b/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
@@ -171,20 +171,6 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 		settingsXML =
 			settingsXML || path.resolve(path.join(os.homedir(), ".m2", "settings.xml"));
 
-		if (!skipConfirmation) {
-			const {default: yesno} = await import("yesno");
-			const ok = await yesno({
-				question:
-					"SnapshotURL is missing. " +
-					`Would you like to try to resolve it from ${settingsXML}? (yes)`,
-				defaultValue: true,
-			});
-
-			if (!ok) {
-				throw new Error("SnapshotURL not resolved");
-			}
-		}
-
 		const {default: fs} = await import("graceful-fs");
 		const {promisify} = await import("node:util");
 		const readFile = promisify(fs.readFile);
@@ -199,20 +185,40 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 			const fileContent = await readFile(settingsXML);
 			xmlContents = await parser.parseStringPromise(fileContent);
 		} catch (err) {
+			skipConfirmation = true;
 			if (err.code === "ENOENT") {
 				// "File or directory does not exist"
-				throw new Error(`SnapshotURL not resolved. Settings.xml could not be found in ${settingsXML}`);
+				log.verbose(`SnapshotURL not resolved. Settings.xml could not be found in ${settingsXML}`);
 			} else {
-				throw new Error(
-					`Failed to read Maven configuration file from ${settingsXML}: ${err.message}`, {
-						cause: err
-					});
+				log.warning(`Failed to read Maven configuration file from ${settingsXML}: ${err.message}`);
 			}
 		}
 
-		return xmlContents?.settings?.profiles[0]?.profile.filter(
+		let url = xmlContents?.settings?.profiles[0]?.profile.filter(
 			(prof) => prof.id[0]._ === "snapshot.build"
 		)[0].pluginRepositories[0].pluginRepository[0].url[0]._;
+
+		if (!url) {
+			skipConfirmation = true;
+			log.verbose(`"snapshot.build" could not be found in ${settingsXML}`);
+		}
+
+		if (!skipConfirmation) {
+			const {default: yesno} = await import("yesno");
+			const ok = await yesno({
+				question:
+					"SnapshotURL is missing. " +
+					`Would you like to try to resolve it from ${settingsXML}? (yes)`,
+				defaultValue: true,
+			});
+
+			if (!ok) {
+				log.verbose(`SnapshotURL not resolved`);
+				url = null;
+			}
+		}
+
+		return url;
 	}
 }
 
--- a/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
+++ b/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
@@ -167,7 +167,7 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 			const ok = await yesno({
 				question:
 					"SnapshotURL is missing. " +
-					"Would you like to try to resolve it from ~/.m2/settings.xml? (yes)",
+					`Would you like to try to resolve it from ${settingsXML}? (yes)`,
 				defaultValue: true,
 			});
 
--- a/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
+++ b/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
@@ -54,7 +54,7 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 		this._installer = new Installer({
 			ui5HomeDir: this._ui5HomeDir,
 			snapshotEndpointUrlCb:
-				Sapui5MavenSnapshotResolver.createSnapshotEndpointUrlCallback(options.snapshotEndpointUrl),
+				Sapui5MavenSnapshotResolver._createSnapshotEndpointUrlCallback(options.snapshotEndpointUrl),
 			cacheMode,
 			artifactsDir,
 			packagesDir,
@@ -136,7 +136,7 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 			ui5HomeDir: path.resolve(
 				ui5HomeDir || path.join(os.homedir(), ".ui5")
 			),
-			snapshotEndpointUrlCb: Sapui5MavenSnapshotResolver.createSnapshotEndpointUrlCallback(snapshotEndpointUrl),
+			snapshotEndpointUrlCb: Sapui5MavenSnapshotResolver._createSnapshotEndpointUrlCallback(snapshotEndpointUrl),
 		});
 		return await installer.fetchPackageVersions({
 			groupId: DIST_GROUP_ID,
@@ -144,7 +144,7 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 		});
 	}
 
-	static createSnapshotEndpointUrlCallback(snapshotEndpointUrl) {
+	static _createSnapshotEndpointUrlCallback(snapshotEndpointUrl) {
 		snapshotEndpointUrl = snapshotEndpointUrl || process.env.UI5_MAVEN_SNAPSHOT_ENDPOINT;
 
 		if (!snapshotEndpointUrl) {
--- a/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
+++ b/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
@@ -13,8 +13,8 @@ const DIST_ARTIFACT_ID = "sapui5-sdk-dist";
  * Resolver for the SAPUI5 framework
  *
  * This Resolver downloads and installs SNAPSHOTS of UI5 libraries from
- * a Maven repository. It's meant for internal usage only as no usecases
- * were forseen out of SAP.
+ * a Maven repository. It's meant for internal usage only as no use cases
+ * outside of SAP are known.
  *
  * @public
  * @class
--- a/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
+++ b/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
@@ -196,7 +196,10 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 				// "File or directory does not exist"
 				throw new Error(`SnapshotURL not resolved. Settings.xml could not be found in ${settingsXML}`);
 			} else {
-				throw err;
+				throw new Error(
+					`Failed to read Maven configuration file from ${settingsXML}: ${err.message}`, {
+					cause: err
+				});
 			}
 		}
 
--- a/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
+++ b/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
@@ -179,41 +179,47 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 			preserveChildrenOrder: true,
 			xmlns: true,
 		});
-		let xmlContents;
+		let url;
 
+		log.verbose(`Attempting to resolve snapshot endpoint URL from Maven configuration file at ${settingsXML}...`);
 		try {
 			const fileContent = await readFile(settingsXML);
-			xmlContents = await parser.parseStringPromise(fileContent);
+			const xmlContents = await parser.parseStringPromise(fileContent);
+
+			url = xmlContents?.settings?.profiles[0]?.profile.filter(
+				(prof) => prof.id[0]._ === "snapshot.build"
+			)[0].pluginRepositories[0].pluginRepository[0].url[0]._;
+
+			if (!url) {
+				skipConfirmation = true;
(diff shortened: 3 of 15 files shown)


lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,121 @@
...
+import Installer from "./maven/Installer.js";
+import {getLogger} from "@ui5/logger";
+const log = getLogger("ui5Framework:Sapui5MavenSnapshotResolver");
+
+const DIST_PKG_NAME = "@sapui5/distribution-metadata";
+const DIST_GROUP_ID = "com.sap.ui5.dist";
+const DIST_ARTIFACT_ID = "sapui5-sdk-dist";
+
+/**
+ * Resolver for the SAPUI5 framework
```
flovogt, line 13: does this resolver has to be public? Its a completly sap internally used feature, isn't?

lib/config/Configuration.js
```diff
@@ -0,0 +1,319 @@
...
+import os from "node:os";
+
+/**
+ *
+ *
+ * @public
+ * @class
+ * @alias @ui5/project/config/Configuration
+ */
+export class Configuration {
```
flovogt, line 11: IMHO we could put the configuration feature in a separate PR, merge that one first and the rebase the snapshot PR.

flovogt: With that we make this PR here not even larger.

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,121 @@
...
+			throw new Error(`Could not find library "${libraryName}"`);
+		}
+
+		return metadata;
+	}
+	async handleLibrary(libraryName) {
+		const metadata = await this.getLibraryMetadata(libraryName);
+		if (!metadata.gav) {
+			throw new Error(
+				`Metadata is missing GAV information. This might indicate an unsupported SNAPSHOT version.`);
```
flovogt, line 82: whats `GAV`?

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,497 @@
+import path from "node:path";
+import {mkdirp} from "../../utils/fs.js";
+import fs from "graceful-fs";
+import _StreamZip from "node-stream-zip";
+const StreamZip = _StreamZip.async;
+import {promisify} from "node:util";
+import Registry from "./Registry.js";
+import AbstractInstaller from "../AbstractInstaller.js";
+import _rimraf from "rimraf";
+const rimraf = promisify(_rimraf);
```
flovogt, line 10: rimraf returns promises instead of callbacks with version 4, so this promisify is not needed.

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,497 @@
...
+
+class Installer extends AbstractInstaller {
+	/**
+	 * @param {object} parameters Parameters
+	 * @param {string} parameters.cwd
+	 * @param {string} parameters.ui5HomeDir UI5 home directory location. This will be used to store packages,
+	 * metadata and configuration used by the resolvers.
+	 * @param {string} parameters.snapshotEndpointUrl Maven repository URL.
+	 * 	Example: <code>https://registry.corp/vendor/build-snapshots/</code>
+	 * @param {string} [parameters.cacheMode="default"] Can be "default" (cache everything, invalidate after 9 hours),
```
flovogt, line 30: might be worth to create an enum for `cacheMode` instead of string and explanation

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,497 @@
...
+	 * 	"off" (do not cache), "force" (use cache only - no requests) and
+	 *  "relaxed" (try to refresh, but fallback to cache if refresh fails for up to one week)
+	 * @param {string} parameters.artifactsDir
+	 * @param {string} parameters.packagesDir
```
flovogt, line 34: missing jsdoc

flovogt: these values are defaulted in the method, so jsdoc should show that

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,497 @@
...
+				// Check whether staging dir already exists and remove it
+				if (await this._pathExists(stagingPath)) {
+					log.verbose(`Removing existing file in staging dir at ${stagingPath}...`);
+					await rm(stagingPath);
+				}
+				await mkdirp(path.dirname(stagingPath));
+
+				log.verbose(`Installing ${artifactId} in version ${version} to ${stagingPath}...`);
+
+				// TODO: Stream response body to installPackage and unzip directly via
```
flovogt, line 392: will we handle this TODO in the initial version?

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,497 @@
...
+		}
+		return {
+			artifactPath: targetPath,
+			removeArtifact: () => {
+				return rm(targetPath);
+			}
+		};
+	}
+
+	async _projectExists(targetDir) {
```
flovogt, line 413: in general we can use private functions with `#` here instead of putting only the `_` in the function name.

lib/ui5Framework/maven/Registry.js
```diff
@@ -0,0 +1,105 @@
...
+import fetch from "make-fetch-happen";
+import xml2js from "xml2js";
+import {pipeline} from "node:stream";
+import {promisify} from "node:util";
+const streamPipeline = promisify(pipeline);
+import fs from "graceful-fs";
+const log = getLogger("ui5Framework:maven:Registry");
+
+class Registry {
+	constructor({cwd, endpointUrl}) {
```
flovogt, line 11: missing jsdoc

lib/ui5Framework/maven/Registry.js
```diff
@@ -0,0 +1,105 @@
...
+		this._endpointUrl = endpointUrl;
+		if (!this._endpointUrl.endsWith("/")) {
+			this._endpointUrl += "/";
+		}
+	}
+
+	/**
+	 * Requests a <code>maven-metadata.xml</code> file from the repository
+	 *
+	 * @param {*} options options
```
flovogt, line 28:
```suggestion
	 * @param {object} options options
```

lib/ui5Framework/maven/Registry.js
```diff
@@ -0,0 +1,105 @@
...
+					`Empty or unexpected response body:\n${content}\nParsed as:\n${JSON.stringify(parsedXml)}`);
+			}
+			return parsedXml.metadata;
+		} catch (err) {
+			if (err.code === "ENOTFOUND") {
+				throw new Error(
+					`Failed to connect to Maven registry at ${this._endpointUrl}. ` +
+					`Please check the correct Endpoint URL is maintained and can be reached. `+
+					`You may be able to continue working offline. For this, set --cache-mode to "force" or ` +
+					`use the --offline flag`); // TODO: Implement --offline flag
```
flovogt, line 66: will this todo be part of initial release? If not, the error message should be adjusted

lib/ui5Framework/maven/Registry.js
```diff
@@ -0,0 +1,105 @@
...
+
+			// Write to target
+			await streamPipeline(res.body, fs.createWriteStream(targetPath));
+		} catch (err) {
+			if (err.code === "ENOTFOUND") {
+				throw new Error(
+					`Failed to connect to Maven registry at ${this._endpointUrl}. ` +
+					`Please check the correct Endpoint URL is maintained and can be reached. `+
+					`You may be able to continue working offline. For this, set --cache-mode to "force" or ` +
+					`use the --offline flag`); // TODO: Implement --offline flag
```
flovogt, line 97: will this todo be part of initial release? If not, the error message should be adjusted

lib/ui5Framework/npm/Installer.js
```diff
@@ -1,29 +1,28 @@
 import path from "node:path";
+import {mkdirp} from "../../utils/fs.js";
 import fs from "graceful-fs";
 import {promisify} from "node:util";
 import Registry from "./Registry.js";
-import rimraf from "rimraf";
+import AbstractInstaller from "../AbstractInstaller.js";
+import _rimraf from "rimraf";
+const rimraf = promisify(_rimraf);
```
flovogt, line 8: rimraf already returns a promise with version 4, so no promisify is needed

lib/ui5Framework/npm/Registry.js
```diff
@@ -19,6 +25,9 @@ class Registry {
 	}
 	async requestPackageManifest(pkgName, version) {
 		const {pacote, pacoteOptions} = await this._getPacote();
+		// TODO: Pacote#manifest essentially installs and unpacks a package
```
flovogt, line 28: will this todo be part of this PR?

lib/config/Configuration.js
```diff
@@ -0,0 +1,319 @@
...
+	const {default: fs} = await import("graceful-fs");
+	const {promisify} = await import("node:util");
+	const writeFile = promisify(fs.writeFile);
+
+	return writeFile(filePath, JSON.stringify(config.toJSON()));
+}
+
+
+export async function resolveSnapshotEndpointUrl(settingsXML) {
+	const {default: yesno} = await import("yesno");
```
flovogt, line 282: here we introduce a coding relevant only for cli usage... what happens when someone is using the `ui5-project` as a direct dependency?

package.json
```diff
@@ -128,14 +129,17 @@
 		"js-yaml": "^4.1.0",
 		"libnpmconfig": "^1.2.1",
 		"lockfile": "^1.0.4",
+		"make-fetch-happen": "^11.0.0",
```
flovogt, line 132: can be increased to version `^11.0.3`

flovogt (changes requested):

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
d3xter666: Not quite sure. The argument here is that the rest of the resolvers are also public.

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
d3xter666: It's a maven acronym

> Repositories store artifacts using a set of coordinates: groupId, artifactId, version, and packaging. The GAV coordinate standard is the foundation for Maven's ability to manage dependencies. This set of coordinates is often referred to as a GAV coordinate, which is short for “Group, Artifact, Version coordinate.”

lib/ui5Framework/maven/Installer.js
d3xter666: Thanks!

lib/ui5Framework/npm/Installer.js
d3xter666: Thanks!

lib/config/Configuration.js
d3xter666: This is a static function, that might be used where needed to resolve the snapshotUrl from settings.xml.
I was indeed wondering where this needs to be positioned, but as it's relevant and at this point, it's a bit too early for the CLI implementation. 
Maybe a flag to enable this check would be sufficient. What do you think?

lib/ui5Framework/maven/Installer.js
d3xter666: Done

lib/ui5Framework/maven/Installer.js
d3xter666: Completelly agree with that, but there are some considerations we need to take into account.

The other Installer has been already made that way. Current implementation would become a lot bigger if we do such reffactoring here. Also we'd need to adjust lots of tests. Maybe we could consider such a refactoriing later.

flovogt: would be fine for me.

lib/config/Configuration.js
d3xter666: Now it's been moved to a separate change

lib/ui5Framework/npm/Registry.js
d3xter666: We've decide we'd address those in a separate change

lib/config/Configuration.js
d3xter666: Moved to a separate change: https://github.com/SAP/ui5-project/pull/575

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,132 @@
...
+ * @extends @ui5/project/ui5Framework/AbstractResolver
+ */
+class Sapui5MavenSnapshotResolver extends AbstractResolver {
+	/**
+	 * @param {*} options options
+	 * @param {string} options.snapshotEndpointUrl Maven Repository Snapshot URL
+	 * @param {string} options.version SAPUI5 version to use
+	 * @param {string} [options.sources=false] Whether to install framework libraries as sources or
+	 * pre-built (with build manifest)
+	 * @param {string} [options.cwd=process.cwd()] Working directory to resolve configurations like .npmrc
```
matz3, line 27: This is copied from AbstractResolver, but as npm is not used, the example of `.npmrc` might be confusing.

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,132 @@
...
+				extension: this._sources ? "zip" : "jar"
+			})
+		};
+	}
+	static async fetchAllVersions({ui5HomeDir, cwd} = {}) {
+		const installer = new Installer({
+			cwd: cwd ? path.resolve(cwd) : process.cwd(),
+			ui5HomeDir:
+				ui5HomeDir ? path.resolve(ui5HomeDir) :
+					path.join(os.homedir(), ".ui5")
```
matz3, line 123: Resolving `os.homedir` should also be ensured here. Same as fixed via https://github.com/SAP/ui5-project/pull/588

lib/ui5Framework/maven/Registry.js
```diff
@@ -0,0 +1,110 @@
+import {getLogger} from "@ui5/logger";
+import fetch from "make-fetch-happen";
+import xml2js from "xml2js";
+import {pipeline} from "node:stream";
+import {promisify} from "node:util";
+const streamPipeline = promisify(pipeline);
```
matz3, line 6: There's `node:stream/promises` since Node v15.0.0

lib/ui5Framework/maven/Registry.js
```diff
@@ -0,0 +1,110 @@
...
+				throw new Error(
+					`Empty or unexpected response body:\n${content}\nParsed as:\n${JSON.stringify(parsedXml)}`);
+			}
+			return parsedXml.metadata;
+		} catch (err) {
+			if (err.code === "ENOTFOUND") {
+				throw new Error(
+					`Failed to connect to Maven registry at ${this._endpointUrl}. ` +
+					`Please check the correct Endpoint URL is maintained and can be reached. `+
+					`You may be able to continue working offline. For this, set --cache-mode to "force"`);
```
matz3, line 70: Message needs to be adjusted to actual implementation of cache mode.

matz3 (changes requested):

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
d3xter666: Furthermore, it's passed to the mvn/Installer and then to mvn/Registry. Within the Registry, there's also a check for it and if missing, an Error is thrown. However, in this scenario, it doesn't seem to be needed at all.

Do you think we might remove it entirely?

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
d3xter666: Thanks!

lib/ui5Framework/maven/Registry.js
d3xter666: The message seems to be valid. Please check here: https://github.com/SAP/ui5-project/blob/rebase-2/lib/ui5Framework/maven/Installer.js#L122

The problem is that this mode now cannot be adjusted by the end-user explicitly. We're missing the UI5 CLI part yet.
Please check here:
- https://github.com/SAP/ui5-project/blob/rebase-2/lib/graph/graph.js#L73
- https://github.com/SAP/ui5-project/blob/rebase-2/lib/graph/graph.js#L123
- https://github.com/SAP/ui5-project/blob/rebase-2/lib/graph/graph.js#L171

d3xter666: The message is updated

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
d3xter666: Removed the `cwd` property for this scenario, because it was completely redundant

test/lib/graph/helpers/ui5Framework.js
```diff
@@ -1585,3 +1698,48 @@ test.serial("ProjectProcessor: Project missing in metadata", async (t) => {
 		message: "Failed to find library lib.a in dist packages metadata.json"
 	}, "Threw with expected error message");
 });
+
+test.serial("resolveSnapshotEndpointUrl", async (t) => {
```
matz3, line 1702: How is this test related to the ui5Framework helper? It doesn't use any of its APIs

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,195 @@
...
+		});
+		return await installer.fetchPackageVersions({
+			groupId: DIST_GROUP_ID,
+			artifactId: DIST_ARTIFACT_ID,
+		});
+	}
+	static async resolveSnapshotEndpointUrl(settingsXML, skipConfirmation) {
+		if (!skipConfirmation) {
+			const {default: yesno} = await import("yesno");
+			const ok = await yesno({
```
matz3, line 148: I tried this locally via `ui5 serve` and it didn't throw or wait and then failed at some other place with `this._endpointUrl.endsWith is not a function`.

matz3 (changes requested):

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
d3xter666: Removed the jsdoc anotation. Would that be sufficient?

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -111,22 +126,70 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
...
+			ui5HomeDir: path.resolve(
+				ui5HomeDir || path.join(os.homedir(), ".ui5")
+			),
 		});
 		return await installer.fetchPackageVersions({
 			groupId: DIST_GROUP_ID,
 			artifactId: DIST_ARTIFACT_ID,
 		});
 	}
+	static async resolveSnapshotEndpointUrl(settingsXML, skipConfirmation) {
```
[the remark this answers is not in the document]
d3xter666: Done

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -111,22 +126,70 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
...
+		}
+
+		settingsXML =
+			settingsXML || path.resolve(path.join(os.homedir(), ".m2", "settings.xml"));
+		const {default: fs} = await import("graceful-fs");
+		const {promisify} = await import("node:util");
+		const readFile = promisify(fs.readFile);
+		const xml2js = await import("xml2js");
+		const parser = new xml2js.Parser({
+			// explicitChildren: true,
```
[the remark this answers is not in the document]
d3xter666: Done

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -111,22 +126,70 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
...
+			xmlns: true,
+		});
+		let xmlContents;
+
+		try {
+			const fileContent = await readFile(settingsXML);
+			xmlContents = await parser.parseStringPromise(fileContent);
+		} catch (err) {
+			if (err.code === "ENOENT") {
+				// "File or directory does not exist"
```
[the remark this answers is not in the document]
d3xter666: Done

test/lib/ui5framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,224 @@
...
+
+	process.env.UI5_MAVEN_SNAPSHOT_ENDPOINT = "_SNAPSHOT_URL_";
+
+	t.context.Sapui5MavenSnapshotResolver = await esmock("../../../lib/ui5Framework/Sapui5MavenSnapshotResolver.js", {
+		"../../../lib/ui5Framework/maven/Installer": t.context.InstallerStub
+	});
+});
+
+test.afterEach.always(() => {
+	process.env.UI5_MAVEN_SNAPSHOT_ENDPOINT = null;
```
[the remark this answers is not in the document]
d3xter666: Done

test/lib/graph/helpers/ui5Framework.js
d3xter666: Inderectly thrugh the resolver. However, you're completely correct that it'd be better to move it to the Sapui5MavenSnapshotResolver test

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,195 @@
...
+	 * @param {string} options.snapshotEndpointUrl Maven Repository Snapshot URL
+	 * @param {string} options.version SAPUI5 version to use
+	 * @param {string} [options.sources=false] Whether to install framework libraries as sources or
+	 * pre-built (with build manifest)
+	 * @param {string} [options.cwd=process.cwd()] Current working directory
+	 * @param {string} [options.ui5HomeDir="~/.ui5"] UI5 home directory location. This will be used to store packages,
+	 * metadata and configuration used by the resolvers. Relative to `process.cwd()`
+	 * @param {string} [options.cacheMode="default"] Can be "default" (cache everything, invalidate after 9 hours),
+	 * 	"off" (do not cache) and "force" (use cache only - no requests)
+	 * @param {string} [options.artifactsDir]
```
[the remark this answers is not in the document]
d3xter666: Done

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,495 @@
...
+class Installer extends AbstractInstaller {
+	/**
+	 * @param {object} parameters Parameters
+	 * @param {string} parameters.ui5HomeDir UI5 home directory location. This will be used to store packages,
+	 * metadata and configuration used by the resolvers.
+	 * @param {string} parameters.snapshotEndpointUrl Maven repository URL.
+	 * 	Example: <code>https://registry.corp/vendor/build-snapshots/</code>
+	 * @param {string} [parameters.cacheMode="default"] Can be "default" (cache everything, invalidate after 9 hours),
+	 * 	"off" (do not cache), "force" (use cache only - no requests) and
+	 *  "relaxed" (try to refresh, but fallback to cache if refresh fails for up to one week)
```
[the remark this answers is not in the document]
d3xter666: Thanks! Removed it

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
d3xter666: Now should be fixed

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
matz3: Maybe I should have phrased my comment differently: I'm okay with having it public API. It could be used internally by other teams.

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -142,7 +153,16 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
...
 	}
-	static async resolveSnapshotEndpointUrl(settingsXML, skipConfirmation) {
+	/**
+	 * Tries to detect whether ~/.m2/settings.xml and if so, the snapshot.build URL is extracted from there
+	 *
+	 * @param {string} [settingsXML=~/.m2/settings.xml] Path to the settings.xml.
+	 * 				If not provided, the default location would be used
+	 * @param {boolean} [skipConfirmation=false] Whether to skip the manual user
+	 * 				confirmation in the consle and enforce reading from settings.xml
+	 * @returns {string} The resolved snapshot.build URL from ~/.m2/settings.xml
```
matz3, line 163: `{Promise<string>}`

d3xter666: Thanks!

matz3 (changes requested):

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,203 @@
...
+ * @class
+ * @alias @ui5/project/ui5Framework/Sapui5MavenSnapshotResolver
+ * @extends @ui5/project/ui5Framework/AbstractResolver
+ */
+class Sapui5MavenSnapshotResolver extends AbstractResolver {
+	/**
+	 * @param {*} options options
+	 * @param {string} options.snapshotEndpointUrl Maven Repository Snapshot URL
+	 * @param {string} options.version SAPUI5 version to use
+	 * @param {string} [options.sources=false] Whether to install framework libraries as sources or
```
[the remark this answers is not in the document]
d3xter666: Done

lib/ui5Framework/AbstractResolver.js
```diff
@@ -32,7 +32,9 @@ class AbstractResolver {
 	 * @param {*} options options
 	 * @param {string} [options.version] Framework version to use. When omitted, all libraries need to be available
 	 * via <code>providedLibraryMetadata</code> parameter. Otherwise an error is thrown.
-	 * @param {string} [options.cwd=process.cwd()] Working directory to resolve configurations like .npmrc
+	 * @param {string} [options.sources=false] Whether to install framework libraries as sources or
```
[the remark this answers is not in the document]
d3xter666: Done

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,505 @@
...
+const mvnTimestampRegex = /^(\d{4})(\d\d)(\d\d)(\d\d)(\d\d)(\d\d)$/;
+
+const CACHE_TIME = 32400000; // 9 hours
+
+class Installer extends AbstractInstaller {
+	/**
+	 * @param {object} parameters Parameters
+	 * @param {string} parameters.ui5HomeDir UI5 home directory location. This will be used to store packages,
+	 * metadata and configuration used by the resolvers.
+	 * @param {Promise<string>} parameters.snapshotEndpointUrl Promise, resolving to Maven repository URL.
```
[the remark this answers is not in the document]
d3xter666: Fully agree with that and that was my idea initially. Please take a look at this commit: https://github.com/SAP/ui5-project/pull/570/commits/3bd183b6cb94390b5c2270fb48266dd6830a9827

It works perfectly fine unless you'd like to work in "offline" mode. The issue is that in that case, although the artifacts and packages are available locally, the end user would still face the prompt with the question of whether to resolve the URL from settings.xml. 
The core reasoning is that the cache logic is inside `maven/Registry`, inside `maven/Installer`. Basically, the earliest possible moment to "know" about the cache is inside the `maven/Installer`

lib/graph/helpers/ui5Framework.js
```diff
@@ -297,7 +299,7 @@ export default {
 		let version = options.versionOverride || frameworkVersion;
 
 		if (rootProject.isFrameworkProject() && !version) {
-			// If the root project is a framework project and no framework version is defined,
+			// If the root project is a framework project, and the framework version is not a snapshot version,
```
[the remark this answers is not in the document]
d3xter666: It came from here: https://github.com/SAP/ui5-project/pull/570/commits/420edd6be49964b6b2396b0e08434937c7ff697b, but currently I don't think also makes sense with the workspaces fix. I'll update the text

lib/ui5Framework/maven/Installer.js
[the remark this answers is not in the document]
matz3: Understood.
So after looking into the code again `parameters.snapshotEndpointUrl` is not of type `Promise<string>`. It is an async callback that resolves with the URL, which makes sense to support the described behavior.
I was confused by the documented type as a Promise gives the installer no way to tell when the result is needed. It could only wait for it.

Please adapt the parameter name to make clear that it is an async callback that returns the snapshotEndpointUrl.

matz3 (changes requested):

lib/ui5Framework/maven/Installer.js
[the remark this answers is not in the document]
d3xter666: Done

matz3 (commented):
LGTM. Title / commit message still needs to be adopted + UA review

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,207 @@
...
+			throw new Error(`Could not find library "${libraryName}"`);
+		}
+
+		return metadata;
+	}
+	async handleLibrary(libraryName) {
+		const metadata = await this.getLibraryMetadata(libraryName);
+		if (!metadata.gav) {
+			throw new Error(
+				`Metadata is missing GAV information. This might indicate an unsupported SNAPSHOT version.`
```
KlattG, line 113: Do people know what GAV stands for?

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,207 @@
...
+			),
+			snapshotEndpointUrlCb: this._snapshotEndpointUrlCb,
+		});
+		return await installer.fetchPackageVersions({
+			groupId: DIST_GROUP_ID,
+			artifactId: DIST_ARTIFACT_ID,
+		});
+	}
+	/**
+	 * Tries to detect whether ~/.m2/settings.xml and if so, the snapshot.build URL is extracted from there
```
KlattG, line 154:
```suggestion
	 * Tries to detect whether ~/.m2/settings.xml exist, and if so, whether the snapshot.build URL is extracted from there
```

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,207 @@
...
+	 *
+	 * @param {string} [settingsXML=~/.m2/settings.xml] Path to the settings.xml.
+	 * 				If not provided, the default location would be used
```
KlattG, line 157:
```suggestion
	 * 				If not provided, the default location is used
```

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,207 @@
...
+	 * @param {boolean} [skipConfirmation=false] Whether to skip the manual user
```
KlattG, line 158:
```suggestion
	 * @param {boolean} [skipConfirmation=false] Whether to skip manual user
```

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,207 @@
...
+	 * 				confirmation in the consle and enforce reading from settings.xml
+	 * @returns {Promise<string>} The resolved snapshot.build URL from ~/.m2/settings.xml
+	 */
+	static async _resolveSnapshotEndpointUrl(settingsXML, skipConfirmation) {
+		if (!skipConfirmation) {
+			const {default: yesno} = await import("yesno");
+			const ok = await yesno({
+				question:
+					"SnapshotURL is missing. " +
+					"Do you want to try to resolve it from ~/.m2/settings.xml? (yes)",
```
KlattG, line 168:
```suggestion
					"Would you like to try to resolve it from ~/.m2/settings.xml? (yes)",
```

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,506 @@
...
+const mvnTimestampRegex = /^(\d{4})(\d\d)(\d\d)(\d\d)(\d\d)(\d\d)$/;
+
+const CACHE_TIME = 32400000; // 9 hours
+
+class Installer extends AbstractInstaller {
+	/**
+	 * @param {object} parameters Parameters
+	 * @param {string} parameters.ui5HomeDir UI5 home directory location. This will be used to store packages,
+	 * metadata and configuration used by the resolvers.
+	 * @param {Function} parameters.snapshotEndpointUrlCb Callback that returns Promise<string>,
```
KlattG, line 26:
```suggestion
	 * @param {Function} parameters.snapshotEndpointUrlCb Callback that returns a Promise <string>,
```

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,506 @@
...
+	 * 	resolving to Maven repository URL.
```
KlattG, line 27:
```suggestion
	 * 	resolving to the Maven repository URL.
```

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,506 @@
...
+	 * 	Example: <code>https://registry.corp/vendor/build-snapshots/</code>
+	 * @param {string} [parameters.cacheMode="default"] Can be "default" (cache everything, invalidate after 9 hours),
+	 * 	"off" (do not cache), "force" (use cache only - no requests)
+	 * @param {string} [parameters.artifactsDir="${ui5HomeDir}/framework/artifacts"] Where to install Maven artifacts
+	 * @param {string} [parameters.packagesDir="${ui5HomeDir}/framework/packages"] Where to install packages
+	 * @param {string} [parameters.metadataDir="${ui5HomeDir}/framework/metadata"] Where to store Maven's
+	 * 	artifacts metadata
```
KlattG, line 34:
```suggestion
	 * @param {string} [parameters.metadataDir="${ui5HomeDir}/framework/metadata"] Where to store the
	 * 	metadata for Maven artifacts
```

lib/ui5Framework/maven/Registry.js
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	/**
+	 * Requests a <code>maven-metadata.xml</code> file from the repository
+	 *
+	 * @param {object} options options
+	 * @param {string} options.groupId
+	 * @param {string} options.artifactId
+	 * @param {string} [options.version] If given, the version must be a SNAPSHOT version.
+	 * 	In that case, the resulting metadata will then list all artifact versions
```
KlattG, line 31:
```suggestion
	 * 	In this case, the resulting metadata will list all artifact versions
```

lib/ui5Framework/maven/Registry.js
```diff
@@ -0,0 +1,108 @@
...
+			if (!parsedXml?.metadata) {
+				throw new Error(
+					`Empty or unexpected response body:\n${content}\nParsed as:\n${JSON.stringify(parsedXml)}`);
+			}
+			return parsedXml.metadata;
+		} catch (err) {
+			if (err.code === "ENOTFOUND") {
+				throw new Error(
+					`Failed to connect to Maven registry at ${this._endpointUrl}. ` +
+					`Please check the correct Endpoint URL is maintained and can be reached. `);
```
KlattG, line 63:
```suggestion
					`Please check the correct endpoint URL is maintained and can be reached. `);
```

lib/ui5Framework/maven/Registry.js
```diff
@@ -0,0 +1,108 @@
...
+				throw new Error(`[HTTP Error] ${res.status} ${res.statusText}`);
+			}
+
+			// Write to target
+			await pipeline(res.body, fs.createWriteStream(targetPath));
+		} catch (err) {
+			if (err.code === "ENOTFOUND") {
+				throw new Error(
+					`Failed to connect to Maven registry at ${this._endpointUrl}. ` +
+					`Please check the correct Endpoint URL is maintained and can be reached. `);
```
KlattG, line 96:
```suggestion
					`Please check the correct endpoint URL is maintained and can be reached. `);
```

KlattG (changes requested):

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
d3xter666: I believe the ones that use this resolver should have that knowledge. However, now I'm wondering whether we could somehow give some context or even a link, for example: https://maven.apache.org/guides/mini/guide-naming-conventions.html

d3xter666: Now it should be better

lib/ui5Framework/Sapui5Resolver.js
```diff
@@ -23,13 +23,19 @@ class Sapui5Resolver extends AbstractResolver {
 	 * @param {string} [options.cwd=process.cwd()] Working directory to resolve configurations like .npmrc
 	 * @param {string} [options.ui5HomeDir="~/.ui5"] UI5 home directory location. This will be used to store packages,
 	 * metadata and configuration used by the resolvers. Relative to `process.cwd()`
+	 * @param {string} [options.cacheDir]
```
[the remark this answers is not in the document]
d3xter666: Done

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,209 @@
...
+	 * the snapshot.build URL is extracted from there
+	 *
+	 * @param {string} [settingsXML=~/.m2/settings.xml] Path to the settings.xml.
+	 * 				If not provided, the default location is used
+	 * @param {boolean} [skipConfirmation=false] Whether to skip manual user
+	 * 				confirmation in the consle and enforce reading from settings.xml
+	 * @returns {Promise<string>} The resolved snapshot.build URL from ~/.m2/settings.xml
+	 */
+	static async _resolveSnapshotEndpointUrl(settingsXML, skipConfirmation) {
+		if (!skipConfirmation) {
```
RandomByte, line 165: Maybe this has already been discussed, but I'm wondering whether we should first attempt to read the `settings.xml` file, and then prompt the user whether to use the content.

In case a user does not have a settings.xml file, we would prompt them whether to use the file, only to then fail with an error because the file does not exist. I think we should avoid this.

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,506 @@
...
+		this._metadataDir = metadataDir ?
+			path.resolve(metadataDir) : path.join(ui5HomeDir, "framework", "metadata");
+		this._stagingDir = stagingDir ?
+			path.resolve(stagingDir) : path.join(ui5HomeDir, "framework", "staging");
+
+		this._cacheMode = cacheMode;
+		this._snapshotEndpointUrlCb = snapshotEndpointUrlCb;
+
+		if (!this._snapshotEndpointUrlCb) {
+			throw new Error(`Installer: Missing Snapshot-Endpoint URL`);
```
RandomByte, line 55:
```suggestion
			throw new Error(`Installer: Missing Snapshot-Endpoint URL callback parameter`);
```

lib/ui5Framework/maven/Installer.js
```diff
@@ -0,0 +1,506 @@
...
+	}
+
+	async getRegistry() {
+		if (this._cachedRegistry) {
+			return this._cachedRegistry;
+		}
+		return (this._cachedRegistry = new Promise((resolve, reject) => {
+			this._snapshotEndpointUrlCb().then((snapshotEndpointUrl) => {
+				if (!snapshotEndpointUrl) {
+					reject(new Error(`Installer: Missing Snapshot-Endpoint URL`));
```
RandomByte, line 70:
```suggestion
					reject(new Error(`Installer: Missing or empty Maven repository URL for snapshot consumption. Please configure a URL to use via the command 'ui5 config set snapshotEndpointUrl https://registry.corp/vendor/build-snapshots/'`));
```

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,209 @@
...
+ *
+ * @public
+ * @class
+ * @alias @ui5/project/ui5Framework/Sapui5MavenSnapshotResolver
+ * @extends @ui5/project/ui5Framework/AbstractResolver
+ */
+class Sapui5MavenSnapshotResolver extends AbstractResolver {
+	/**
+	 * @param {*} options options
+	 * @param {string} options.snapshotEndpointUrl Maven Repository Snapshot URL
```
RandomByte, line 27:
```suggestion
	 * @param {string} [options.snapshotEndpointUrl] Maven Repository Snapshot URL. If not provided, falls back to an optional <code>UI5_MAVEN_SNAPSHOT_ENDPOINT</code> environment variable, or the standard Maven settings.xml file (if existing).
```

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,209 @@
...
+			const {default: yesno} = await import("yesno");
+			const ok = await yesno({
+				question:
+					"SnapshotURL is missing. " +
+					"Would you like to try to resolve it from ~/.m2/settings.xml? (yes)",
+				defaultValue: true,
+			});
+
+			if (!ok) {
+				throw new Error("SnapshotURL not resolved");
```
RandomByte, line 175: Rather than throwing an error, we could also return undefined and let the Installer handle the missing URL parameter. What do you think?

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,209 @@
...
+		});
+		let xmlContents;
+
+		try {
+			const fileContent = await readFile(settingsXML);
+			xmlContents = await parser.parseStringPromise(fileContent);
+		} catch (err) {
+			if (err.code === "ENOENT") {
+				// "File or directory does not exist"
+				throw new Error(`SnapshotURL not resolved. Settings.xml could not be found in ${settingsXML}`);
```
RandomByte, line 197: Same here, rather than throwing an error, maybe we just return undefined?

A user without a Maven settings.xml should probably configure the Snapshot URL using `ui5 config` rather than trying to fix the missing settings.xml. So maybe don't bother throwing an error about the missing file?

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,209 @@
...
+			} else {
+				throw err;
```
RandomByte, line 199: Consider adding some context to the error message:

```suggestion
				throw new Error(
					`Failed to read Maven configuration file from ${settingsXML}: ${err.message}`, {
					cause: err
				});
```

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,209 @@
...
+	 * 				confirmation in the consle and enforce reading from settings.xml
+	 * @returns {Promise<string>} The resolved snapshot.build URL from ~/.m2/settings.xml
+	 */
+	static async _resolveSnapshotEndpointUrl(settingsXML, skipConfirmation) {
+		if (!skipConfirmation) {
+			const {default: yesno} = await import("yesno");
+			const ok = await yesno({
+				question:
+					"SnapshotURL is missing. " +
+					"Would you like to try to resolve it from ~/.m2/settings.xml? (yes)",
```
RandomByte, line 170: Take the settingsXML parameter into account. Move the defaulting above this line.

```suggestion
					`Would you like to try to resolve it from ${settingsXML}? (yes)`,
```

lib/ui5Framework/maven/Registry.js
```diff
@@ -0,0 +1,108 @@
...
+	 * @param {string} parameters.endpointUrl Maven's endpoint URL
+	 */
+	constructor({endpointUrl}) {
+		if (!endpointUrl) {
+			throw new Error(`Registry: Missing parameter "endpointUrl"`);
+		}
+		this._endpointUrl = endpointUrl;
+		if (!this._endpointUrl.endsWith("/")) {
+			this._endpointUrl += "/";
+		}
```
RandomByte, line 21: For error analysis purposes, we should add a verbose log for the effective endpointUrl value

RandomByte (changes requested):

lib/ui5Framework/maven/Registry.js
d3xter666: Done

KlattG (dismissed):
+1

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,209 @@
...
+			cacheMode,
+			artifactsDir,
+			packagesDir,
+			metadataDir,
+			stagingDir,
+			sources,
+		} = options;
+
+		if (!snapshotEndpointUrl) {
+			this._snapshotEndpointUrlCb = Sapui5MavenSnapshotResolver._resolveSnapshotEndpointUrl;
```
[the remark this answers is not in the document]
d3xter666: Done

RandomByte:
Hi @KlattG, I updated some user-facing texts and would appreciate your feedback. Here's the delta: [`8e3f4c4` (#570)](https://github.com/SAP/ui5-project/pull/570/commits/8e3f4c4f044f0829bdde59e251fba5caf753521e)

RandomByte (approved):
LGTM

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,235 @@
...
+			ui5HomeDir: this._ui5HomeDir,
+			snapshotEndpointUrlCb: this._snapshotEndpointUrlCb,
+			cacheMode,
+			artifactsDir,
+			packagesDir,
+			metadataDir,
+			stagingDir,
+		});
+		this._loadDistMetadata = null;
+		this._sources = !!sources;
```
matz3, line 79: `this._sources` is already set by the AbstractResolver

d3xter666: Thanks

KlattG (approved):
LGTM

RandomByte (approved):

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -0,0 +1,233 @@
...
+			}),
+		};
+	}
+	static async fetchAllVersions({ui5HomeDir, cwd} = {}) {
+		const installer = new Installer({
+			cwd: cwd ? path.resolve(cwd) : process.cwd(),
+			ui5HomeDir: path.resolve(
+				ui5HomeDir || path.join(os.homedir(), ".ui5")
+			),
+			snapshotEndpointUrlCb: this._snapshotEndpointUrlCb,
```
RandomByte, line 151: Since this method is static, `this._snapshotEndpointUrlCb` is undefined and needs to be provided as a parameter.

I noticed this only in a manual test locally. I'll add a commit fixing this

RandomByte (changes requested):

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -208,6 +208,13 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 			}
 		}
 
+		if (!process.stdout.isTTY) {
+			// Do not prompt if stdout is non-interactive (i.e. in CI environments) and just use the URL
```
matz3, line 212: Isn't this against the idea to first ask for confirmation to use the URL?

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -144,30 +129,45 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
...
-			snapshotEndpointUrlCb: this._snapshotEndpointUrlCb,
+			snapshotEndpointUrlCb: Sapui5MavenSnapshotResolver.createSnapshotEndpointUrlCallback(snapshotEndpointUrl),
 		});
 		return await installer.fetchPackageVersions({
 			groupId: DIST_GROUP_ID,
 			artifactId: DIST_ARTIFACT_ID,
 		});
 	}
+
+	static createSnapshotEndpointUrlCallback(snapshotEndpointUrl) {
```
matz3, line 147: Prefix with underscore to mark as private API

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
RandomByte: Yeah, I was asking myself that too. Since we can't ask the user in this case, we need to decide upfront.

I went with using the Maven config since I expect this to make things easier in CI environments where the config is already present (I know of some base images where this is the case). But you're right, it might be unwanted or unexpected.

However, the only alternative I see is to always ignore the Maven configuration in CI environments, forcing users to configure the environment variable or ui5 config.

lib/ui5Framework/maven/Registry.js
RandomByte: I hope we'll have it as part of the initial release of the feature

PR #574, opened 2023-02-13
[FIX] Project#getReader: Do not apply builder resource excludes for style 'runtime'

RandomByte:
Introduce new style `dist` to reflect the standard build output
structure which can be used with any HTTP server. It is identical to the
`runtime` style, but also applies any configured [builder resource
excludes](https://sap.github.io/ui5-tooling/stable/pages/Configuration/#exclude-resources).

ProjectBuilder now uses style 'dist' instead of 'runtime'.

Fixes https://github.com/SAP/ui5-tooling/issues/779


matz3 (approved):

Magoli1:
@RandomByte Any ETA for releasing a new version with this change? 🚀

RandomByte:
@Magoli1 Probably later today!

PR #575, opened 2023-02-13
[FEATURE] Add Configuration

RandomByte:
JIRA: CPOUI5FOUNDATION-634

Complements: https://github.com/SAP/ui5-project/pull/570


lib/config/Configuration.js
```diff
@@ -0,0 +1,332 @@
...
+
+/**
+ * Save Configuration to a JSON file
+ *
+ * @public
+ * @static
+ * @param {string} [filePath="~/.ui5rc"] Path to configuration JSON file
+ * @param {@ui5/project/config/Configuration} config Configuration to save
+ */
+export async function saveConfig(filePath, config) {
```
d3xter666, line 279: It's clear from the [RFC](https://github.com/SAP/ui5-tooling/blob/rfc-configuration/rfcs/0013-configuration.md#changing-configuration) that this should be part of the CLI.

Would be removed.

lib/config/Configuration.js
```diff
@@ -0,0 +1,332 @@
...
+
+	const {default: fs} = await import("graceful-fs");
+	const {promisify} = await import("node:util");
+	const writeFile = promisify(fs.writeFile);
+
+	return writeFile(filePath, JSON.stringify(config.toJSON()));
+}
+
+
+export async function resolveSnapshotEndpointUrl(settingsXML, skipConfirmation) {
```
d3xter666, line 290: We've discussed this and need to look at it from another perspective- the end consumer i.e. the developers.
We do not expect them to have Maven installed. 
With the potential CLI implementation of the configuration settings, addressed here: https://github.com/SAP/ui5-tooling/blob/rfc-configuration/rfcs/0013-configuration.md#changing-configuration, it might become even redundant. Developers, when setting their environment, would eventually need just to execute once a single CLI command like `ui5 config set snapshotUrl https://.....` 

If we decide that we still need this revolver, its place must be somewhere in the mvn/Installer where it actually be in use.

lib/config/Configuration.js
```diff
@@ -0,0 +1,332 @@
...
+		stagingDir,
+		lockDir
+	}) {
+		this.#cwd = cwd ? path.resolve(cwd) : process.cwd();
+		this.#sources = !!sources;
+		this.#version = version;
+		this.#cacheMode = cacheMode;
+		this.#ui5HomeDir = ui5HomeDir ? path.resolve(ui5HomeDir) : path.join(os.homedir(), ".ui5");
+
+		this.#snapshotEndpointUrl = process.env.UI5_MAVEN_SNAPSHOT_ENDPOINT || snapshotEndpointUrl;
```
d3xter666, line 73: The env variable should not be part of the Configuration object.

lib/config/Configuration.js
```diff
@@ -0,0 +1,332 @@
...
+		this.#frameworkDir = frameworkDir ? path.resolve(frameworkDir) : path.join(this.#ui5HomeDir, "framework");
+		this.#artifactsDir = artifactsDir ?
+			path.resolve(artifactsDir) : path.join(this.#frameworkDir, "artifacts");
+		this.#packagesDir = packagesDir ?
+			path.resolve(packagesDir) : path.join(this.#frameworkDir, "packages");
+		this.#metadataDir = metadataDir ?
+			path.resolve(metadataDir) : path.join(this.#frameworkDir, "metadata");
+		this.#stagingDir = stagingDir ? path.resolve(stagingDir) : path.join(this.#frameworkDir, "staging");
+		this.#lockDir = lockDir ? path.resolve(lockDir) : path.join(this.#frameworkDir, "locks");
+		this.#cacheDir = cacheDir ? path.resolve(cacheDir) : path.join(this.#frameworkDir, "cacache");
```
d3xter666, line 83: There are several questions that arise here.

### Which properties/settings should be within the Configuration object?

Currently, all the needed configuration options are handled here. The reasoning is that many depend on each other i.e. `frameworkDir` depends on `ui5HomeDir`

On the other side, the core reasoning to introduce such a Configuration was to support the Snapshot URL.

### Do we need default values?

According to the (spec)[https://github.com/SAP/ui5-tooling/blob/rfc-configuration/rfcs/0013-configuration.md], there shouldn't be default values.

Properties could be defined/provided separately. But the way we use them internally implies that they depend on each other i.e. `frameworkDir` depends on `ui5HomeDir`

d3xter666: For this iteration, we have decided to go only with the necessary `snapshotEndpointUrl`. The questions above would be resolved in separate iteration(s)/BLI(s).

lib/config/Configuration.js
d3xter666: Given the latest discussions and the future integration with Snapshot Consumption, it's worth it that the Config object would be able to read and write into the `.ui5rc` file. 

The core reasoning behind that is the following scenario:

1. There's an exported static function that reads the file, creates and returns the Configuration object.

2. Writing in `.ui5rc` is a bit more complex: if no `.ui5rc` config is provided, then the Resolver checks for env variable, but there's a last resort if none of these is provided (new project setup, for example). In such a case, it's worth checking for `settings.xml` in case the end-user has maven installed on their machine. Then we need to read the `settings.xml`, extract the URL and save it to `.ui5rc`

We have decided to give it a try. Furthermore, the Configuration object is closed for now and we could eventually revise it

lib/config/Configuration.js
```diff
@@ -0,0 +1,91 @@
...
+ * @public
+ * @class
+ * @alias @ui5/project/config/Configuration
+ */
+export class Configuration {
+	#snapshotEndpointUrl;
+
+	/**
+	 * @param {object} configuration
+	 * @param {string} [configuration.snapshotEndpointUrl=""]
```
matz3, line 16: There is no defaulting to an empty string in here. So the JSDoc should be updated to not list a default value.

lib/config/Configuration.js
```diff
@@ -0,0 +1,91 @@
...
+
+export default Configuration;
+
+/**
+ * Create Configuration from JSON file
+ *
+ * @public
+ * @static
+ * @param {string} [filePath="~/.ui5rc"] Path to configuration JSON file
+ * @returns {@ui5/project/config/Configuration} Configuration instance
```
matz3, line 51: Promise

lib/config/Configuration.js
d3xter666: Done

lib/config/Configuration.js
d3xter666: Thanks!

lib/config/Configuration.js
```diff
@@ -0,0 +1,91 @@
...
+import os from "node:os";
+
+/**
+ *
+ *
+ * @public
+ * @class
+ * @alias @ui5/project/config/Configuration
+ */
+export class Configuration {
```
matz3, line 11: Remove export. Class is already the default export. This should also resolve the issue of missing methods in JSDoc.

lib/config/Configuration.js
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Create Configuration from JSON file
+ *
+ * @public
+ * @static
+ * @param {string} [filePath="~/.ui5rc"] Path to configuration JSON file
+ * @returns {Promise<@ui5/project/config/Configuration>} Configuration instance
+ */
+export async function fromFile(filePath) {
```
matz3, line 53: fromFile and saveConfig are currently not visible in JSDoc.
As the file represents the Configuration class (as it's written with uppercase "C") it would be best to only export the class.
What about adding those two functions as static methods to the class?

matz3 (changes requested):

lib/config/Configuration.js
d3xter666: Done

lib/config/Configuration.js
d3xter666: Done

lib/config/Configuration.js
```diff
@@ -0,0 +1,91 @@
...
+ * @public
+ * @class
+ * @alias @ui5/project/config/Configuration
+ */
+export class Configuration {
+	#snapshotEndpointUrl;
+
+	/**
+	 * @param {object} configuration
+	 * @param {string} [configuration.snapshotEndpointUrl]
```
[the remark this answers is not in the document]
d3xter666: Done

lib/config/Configuration.js
```diff
@@ -0,0 +1,91 @@
...
+	 */
+	constructor({snapshotEndpointUrl}) {
+		this.#snapshotEndpointUrl = snapshotEndpointUrl;
+	}
+
+	/**
+	 * SNAPSHOTs URL
```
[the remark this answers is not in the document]
d3xter666: Now should be better

matz3 (dismissed):

lib/config/Configuration.js
```diff
@@ -0,0 +1,91 @@
...
+		}
+		return new Configuration(config);
+	}
+
+	/**
+	 * Save Configuration to a JSON file
+	 *
+	 * @public
+	 * @static
+	 * @param {string} [filePath="~/.ui5rc"] Path to configuration JSON file
```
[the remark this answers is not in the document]
d3xter666: Ah, thanks a lot for pointing it out!

lib/config/Configuration.js
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Provides basic configuration settings for the resolvers.
+ * Reads/writes configuration from/to ~/.ui5rc
+ *
+ * @public
+ * @class
+ * @alias @ui5/project/config/Configuration
+ */
+export default class Configuration {
```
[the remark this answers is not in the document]
d3xter666: Sorry, didn't pay attention to that. Now is fixed

test/lib/config/Configuration.js
```diff
@@ -0,0 +1,122 @@
...
+});
+
+test.serial("saveConfig", async (t) => {
+	const {promisifyStub, sinon, Configuration} = t.context;
+	const saveConfig = Configuration.saveConfig;
+
+	const writeStub = sinon.stub().resolves();
+	promisifyStub.callsFake(() => writeStub);
+
+	const config = new Configuration({});
```
RandomByte, line 114: Maybe we could initialize the config with a value for `snapshotEndpointUrl`?

RandomByte (commented):
LGTM, just a small remark

test/lib/config/Configuration.js
d3xter666: Done

lib/config/Configuration.js
```diff
@@ -0,0 +1,93 @@
...
+	/**
+	 * Saves Configuration to a JSON file
+	 *
+	 * @public
+	 * @static
+	 * @param {@ui5/project/config/Configuration} config Configuration to save
+	 * @param {string} [filePath="~/.ui5rc"] Path to configuration JSON file
+	 * @returns {Promise<void>}
+	 */
+	static async saveConfig(config, filePath) {
```
RandomByte, line 82:
```suggestion
	static async toFile(config, filePath) {
```

RandomByte:
LGTM, but I can't approve since I'm the author of the PR 😅

matz3 (dismissed):
LGTM but I wonder whether the changes in `package-lock.json` are related to this PR as no dependencies have been introduced, right?

RandomByte:
Rebased to resolve conflict with main

matz3 (approved):

PR #579, opened 2023-02-14
[FIX] Prevent socket timeouts when installing framework libraries

matz3:


lib/ui5Framework/npm/Registry.js
```diff
@@ -39,7 +39,13 @@ class Registry {
 		if (!this._npmConfig) {
 			const {default: libnpmconfig} = await import("libnpmconfig");
 			const opts = {
-				cache: this._cacheDir
+				cache: this._cacheDir,
```
flovogt, line 42: any test for that?

matz3: We currently don't have real e2e test. One thing we could at least test is whether the expected options are passed to pacote.

flovogt: That would be fine.

flovogt (approved):

PR #580, opened 2023-02-15
[FIX] TaskUtil: Provide framework configuration getters to custom tasks

RandomByte:
This was missed to add to the initial 3.0.0 release

### DIFF
--- a/lib/build/helpers/TaskUtil.js
+++ b/lib/build/helpers/TaskUtil.js
@@ -195,6 +195,9 @@ class TaskUtil {
 	 * @property {Function} getSourcePath Get the local File System path of the project's source directory
 	 * @property {Function} getCustomConfiguration Get the project Custom Configuration
 	 * @property {Function} isFrameworkProject Check whether the project is a UI5-Framework project
+	 * @property {Function} getFrameworkName Get the project's frameworkName configuration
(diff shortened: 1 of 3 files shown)


d3xter666 (commented):
LGTM

lib/build/helpers/TaskUtil.js
```diff
@@ -195,6 +195,9 @@ class TaskUtil {
 	 * @property {Function} getSourcePath Get the local File System path of the project's source directory
 	 * @property {Function} getCustomConfiguration Get the project Custom Configuration
 	 * @property {Function} isFrameworkProject Check whether the project is a UI5-Framework project
+	 * @property {Function} getFrameworkName Get the project's frameworkName configuration
```
flovogt, line 198: above we have used the syntax `framework project`. Should write here `Get the project's framework name`. Same for the other new methods?

flovogt (changes requested):

lib/build/helpers/TaskUtil.js
RandomByte: Yes, but I would keep the "configuration" to make it clear that it's nothing implicit (like for the isFrameworkProject check)

flovogt (approved):

PR #582, opened 2023-02-16
[FIX] ComponentProject#getWorkspace: Apply builder resource excludes

RandomByte:
This resolves a regression caused by
https://github.com/SAP/ui5-project/pull/574 where the application of
builder resource-exclude configuration was moved from the low-level
_getSourceReader/_getTestReader methods to the high-level getReader API.

However, it was missed to also adapt the high-level getWorkspace API in
the same manner.

This caused the created workspace reader/writer to ignore any exclude
configuration as reported in
https://github.com/SAP/ui5-tooling/issues/784#issuecomment-1433625146


codeworrior (approved):
LGTM

PR #588, opened 2023-03-07
[FIX] Resolve properly absolute path for ui5HomeDir

d3xter666:
This fix resolves https://github.com/SAP/ui5-tooling/issues/796

The core issue is that in particular CI environments `os.homedir()` might be enforced to return a relative path.
We rely only on absolute paths in the tooling, so we'd need to ensure this.

### DIFF
--- a/lib/ui5Framework/AbstractResolver.js
+++ b/lib/ui5Framework/AbstractResolver.js
@@ -43,11 +43,11 @@ class AbstractResolver {
 			throw new Error(`AbstractResolver: Missing parameter "version"`);
 		}
 
-		this._ui5HomeDir = ui5HomeDir ?
-			path.resolve(ui5HomeDir) :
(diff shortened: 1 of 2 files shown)


lib/ui5Framework/AbstractResolver.js
```diff
@@ -43,7 +43,11 @@ class AbstractResolver {
 			throw new Error(`AbstractResolver: Missing parameter "version"`);
 		}
 
-		this._ui5HomeDir = ui5HomeDir ? path.resolve(ui5HomeDir) : path.join(os.homedir(), ".ui5");
+		this._ui5HomeDir = ui5HomeDir ?
```
matz3, line 46: This assignment can be simplified a lot, as path.resolve should always be called now.

matz3 (changes requested):

matz3 (approved):

PR #589, opened 2023-03-10
[FIX] ui5Framework: Prevent install of libraries within workspace

matz3:

PR #592, opened 2023-03-29
[FEATURE] Workspace: Add getModules method / Specification: Add getId method

matz3:
Required for https://github.com/SAP/ui5-project/pull/589

### DIFF
--- a/lib/graph/Workspace.js
+++ b/lib/graph/Workspace.js
@@ -96,7 +96,7 @@ class Workspace {
 	 * @public
 	 * @param {string} projectName Name of the project
 	 * @returns {Promise<@ui5/project/graph/Module|undefined>}
-	 *   Module instance or <code>undefined</code> if none is found
+	 *   Module instance, or <code>undefined</code> if none is found
 	 */
 	async getModuleByProjectName(projectName) {
 		const {projectNameMap} = await this._getResolvedModules();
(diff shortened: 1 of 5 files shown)


flovogt (commented):
LGTM. Waiting for review by @KlattG

lib/graph/Workspace.js
```diff
@@ -74,6 +74,20 @@ class Workspace {
...
+	 * Returns an array of [Module]{@ui5/project/graph/Module} instances found in the configured
+	 * dependency-management resolution paths of this workspace, sorted by module ID.
+	 *
+	 * @public
+	 * @returns {Promise<@ui5/project/graph/Module[]>}
+	 *   Array of Module instances sorted by module ID
+	 */
+	async getModules() {
+		const {moduleIdMap} = await this._getResolvedModules();
+		const sortedMap = new Map([...moduleIdMap].sort((a, b) => String(a[0]).localeCompare(b[0])));
```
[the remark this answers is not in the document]
matz3: No, as it will be casted to a string by localeCompare. I added the first `String()` to make sure that localeCompare can be called. But we could also omit that as we know that we only use strings as keys.

KlattG (changes requested):

lib/specifications/Specification.js
```diff
@@ -177,7 +195,7 @@ class Specification {
 	}
 
 	/**
-	 * Get the type of this specification,
+	 * Gest the type of this specification,
```
KlattG, line 198:
```suggestion
	 * Gets the type of this specification,
```

KlattG (changes requested):
Typo

lib/specifications/Specification.js
matz3: Oops. Thanks.

KlattG (approved):
LGTM

PR #599, opened 2023-04-19
[INTERNAL] AbstractResolver: Add env var to enforce usage of framework lib sources

matz3:
By setting environment variable 'UI5_PROJECT_USE_FRAMEWORK_SOURCES' to
any value, resolvers will be forced to always download the sources of
framework dependencies rather than pre-built versions.

Currently this only really affects the Sapui5MavenSnapshotResolver,
being the only resolver capable of downloading pre-built packages as of
today.

JIRA: CPOUI5FOUNDATION-657

### DIFF
--- a/lib/ui5Framework/AbstractResolver.js
+++ b/lib/ui5Framework/AbstractResolver.js
@@ -55,7 +55,13 @@ class AbstractResolver {
 		);
 		this._cwd = cwd ? path.resolve(cwd) : process.cwd();
 		this._version = version;
+
+		// Environment variable should always enforce usage of sources
+		if (process.env.UI5_PROJECT_USE_FRAMEWORK_SOURCES) {
+			sources = true;
+		}
 		this._sources = !!sources;
+
 		this._providedLibraryMetadata = providedLibraryMetadata;
 	}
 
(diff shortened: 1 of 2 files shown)


RandomByte:
LGTM

lib/ui5Framework/AbstractResolver.js
```diff
@@ -55,7 +55,13 @@ class AbstractResolver {
 		);
 		this._cwd = cwd ? path.resolve(cwd) : process.cwd();
 		this._version = version;
+
+		// Environment variable should always enforce usage of sources
+		if (process.env.UI5_PROJECT_USE_FRAMEWORK_SOURCES) {
```
d3xter666, line 60: Do we plan to introduce this option as a CLI command, for example?

I think that for the maven snapshot endpoint, the ENV variable was the last one taken into account if all the other options were missed. 
If we plan to be able to set the `sources` anyhow from the outside, in this case the ENV variable would always take the precedence and would always overwrite it.

In this case, `UI5_PROJECT_USE_FRAMEWORK_SOURCES` is the only way to change `sources` value from outside. From that perspective the change LGTM

flovogt (approved):

lib/ui5Framework/AbstractResolver.js
RandomByte: I changed to prioritization for the `UI5_MAVEN_SNAPSHOT_ENDPOINT_URL` via https://github.com/SAP/ui5-project/pull/601

PR #602, opened 2023-04-21
[FIX] Configuration: Rename toJSON => toJson

RandomByte:
As per our naming convention

### DIFF
--- a/lib/config/Configuration.js
+++ b/lib/config/Configuration.js
@@ -35,7 +35,7 @@ class Configuration {
 	 * @public
 	 * @returns {object} The configuration in a JSON format
 	 */
-	toJSON() {
+	toJson() {
 		return {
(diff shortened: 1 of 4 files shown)


flovogt (approved):

matz3 (dismissed):
What about https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify ?

RandomByte:
> What about [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) ?

True. That is something to take into consideration. However, I don't expect third parties to serialize our Configuration instances. Plus, it's an arguably lesser known convenience feature that is certainly not a deal breaker for any scenario if not available. If this is desperately needed we could even add a "hidden" `toJSON` function without JSDoc in the future...

PR #608, opened 2023-05-02
[FIX] Resolve properly package.json dependency aliases

d3xter666:
Resolves: https://github.com/SAP/ui5-tooling/issues/809

When shimmed packages are defined as dependency aliases in the package.json, they are being excluded from the bundlig as the packager eventually finds the real package and its path and as it's not defined in the shim, but its alias, it gets ignored.

We need to provide more robust discovery in order to handle those cases:
- https://github.com/npm/cli/pull/3
- https://github.com/npm/rfcs/blob/main/implemented/0001-package-aliases.md


test/fixtures/application.a.aliases/webapp/index.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Application A</title>
+</head>
+<body>
+
+</body>
+</html>
```
flovogt, line 9: would add here an empty last line

test/fixtures/application.a.aliases/webapp/manifest.json
```diff
@@ -0,0 +1,13 @@
...
+        "_version": "1.1.0",
+        "id": "id1",
+        "type": "application",
+        "applicationVersion": {
+            "version": "1.2.2"
+        },
+        "embeds": ["embedded"],
+        "title": "{{title}}"
+    }
+}
```
flovogt, line 13: would add here an empty last line

test/fixtures/application.a.aliases/webapp/index.html
d3xter666: Done

test/fixtures/application.a.aliases/webapp/manifest.json
d3xter666: Done

lib/graph/providers/NodePackageDependencies.js
```diff
@@ -99,15 +99,15 @@ class NodePackageDependencies {
 		}
 	}
 
-	async _getNode(modulePath, optional) {
+	async _getNode(modulePath, optional, nameAlias) {
```
RandomByte, line 102: I think it would help us in the future to add a JSDoc comment here describing that this refers to the npm name alias concept

RandomByte (changes requested):

test/lib/graph/providers/NodePackageDependencies.integration.js
```diff
@@ -68,6 +69,25 @@ test("AppA: project with collection dependency", async (t) => {
 	]);
 });
 
+test("AppA: project with an alias dependency", async (t) => {
+	const workspace = {
+		getName: () => "workspace name",
+		getModuleByNodeId: t.context.sinon.stub().resolves(undefined).onFirstCall().resolves({
+			// This version of library.d has an additional dependency to library.f,
+			// which in turn has a dependency to library.g
```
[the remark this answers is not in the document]
d3xter666: Indeed. Thanks!

lib/graph/providers/NodePackageDependencies.js
d3xter666: Done

d3xter666: Done

RandomByte (approved):

PR #611, opened 2023-05-17
[FIX] Properly resolve dependent shimmed modules

d3xter666:
Fixes: https://github.com/SAP/ui5-tooling/issues/807

If an external (non-UI5) module is provided as devDependency in an application, but there's a dependecy to a sub-module that uses the external module as a **main** dependency and is shimmed there, it's been cached without the shim and is not build properly later.

Sample structure:
```bash
app/
  \_ node_modules/
    \_ rxjs/ # devDependency, not shimmed
    \_ library/
      \_ node_modules/
        \_ rxjs/ # dependency, shimmed
```


d3xter666:
A better solution has been proposed

PR #612, opened 2023-05-17
[FIX] projectGraphBuilder: Add module cache invalidation

RandomByte:
If a visited module did not resolve to any specification (project or
extensions), try recreating the module when visiting it again from a
different parent.

Resolves https://github.com/SAP/ui5-tooling/issues/807

This is an alternative to https://github.com/SAP/ui5-project/pull/611

This change might have a greater impact on performance since er might
recreate modules that are not relevant to UI5 Tooling more than once if
they are listed multiple times in the dependency tree.
Before this change, such modules where only visited once.

### DIFF
--- a/lib/graph/projectGraphBuilder.js
+++ b/lib/graph/projectGraphBuilder.js
@@ -161,6 +161,24 @@ async function projectGraphBuilder(nodeProvider, workspace) {
 		const {nodes, parentProject} = queue.shift(); // Get and remove first entry from queue
 		const res = await Promise.all(nodes.map(async (node) => {
 			let ui5Module = moduleCollection[node.id];
+
+			if (ui5Module) {
+				log.silly(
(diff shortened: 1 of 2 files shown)


lib/graph/projectGraphBuilder.js
```diff
@@ -174,8 +192,6 @@ async function projectGraphBuilder(nodeProvider, workspace) {
 					shimCollection
 				});
 			} else {
-				log.silly(
-					`Re-visiting module ${node.id} as a dependency of ${parentProject.getName()}`);
 				if (ui5Module.getPath() !== node.path) {
```
matz3, line 195: Can be converted to else-if

matz3 (approved):

lib/graph/projectGraphBuilder.js
RandomByte: Done

PR #622, opened 2023-06-14
[FEATURE] Sapui5MavenSnapshotResolver: Use npm-dist.zip artifact for 1.116.0 and later

RandomByte:
JIRA: CPOUI5FOUNDATION-672

### DIFF
--- a/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
+++ b/lib/ui5Framework/Sapui5MavenSnapshotResolver.js
@@ -108,23 +108,20 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
 			// Use npm-sources artifact if sources are requested
 			classifier = "npm-sources";
 			extension = "zip";
-		} else if (this._isLegacySnapshotVersion) {
-			// For legacy versions < 1.116.0-SNAPSHOT where npm-dist artifact is not
-			// yet available, use "default" JAR
-			classifier = null;
-			extension = "jar";
-
+		} else {
 			// Add "prebuilt" suffix to package name
 			pkgName += "-prebuilt";
-		} else {
-			// Use "npm-dist" artifact by default
-			classifier = "npm-dist";
-			extension = "zip";
 
-			// Add "-prebuilt_dist" suffix to package name
-			// We can't use "-prebuilt" since old ui5-project versions might continue to download
-			// and store the default JAR under that name too
-			pkgName += "-prebuilt_dist";
+			if (this._isLegacySnapshotVersion) {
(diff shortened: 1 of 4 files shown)


lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -95,8 +100,31 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
...
-		if (!this._sources) {
+
+		// Use "npm-dist" artifact by default
+		let classifier;
+		let extension;
+		if (this._sources) {
+			// Use npm-sources artifact if sources are requested
+			classifier = "npm-sources";
+			extension = "zip";
+		} else if (this._isLegacySnapshotVersion) {
```
flovogt, line 111: might add a TODO for Tooling V4 to remove this if again?

flovogt (commented):
LGTM but wait for others review

d3xter666:
LGTM

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
```diff
@@ -95,8 +100,31 @@ class Sapui5MavenSnapshotResolver extends AbstractResolver {
...
 			pkgName += "-prebuilt";
+		} else {
+			// Use "npm-dist" artifact by default
+			classifier = "npm-dist";
+			extension = "zip";
+
+			// Add "-prebuilt_dist" suffix to package name
+			// We can't use "-prebuilt" since old ui5-project versions might continue to download
+			// and store the default JAR under that name too
+			pkgName += "-prebuilt_dist";
```
matz3, line 127: I wonder whether this is really the best solution. This is not the actual package name anymore.
Maybe the installer should rather have the knowledge on how to name the folders, because that's what is important here. The package name is actually not changing.

RandomByte: Matthias and I found a potential solution to this. With the latest commit, we check whether the found "-prebuilt" package actually contains a package.json (which is only the case for the new npm-dist.zip artifact). In case no package.json is found, the installed package is overwritten with a newly installed npm-dist.zip package 👍

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
RandomByte: Done

lib/ui5Framework/Sapui5MavenSnapshotResolver.js
flovogt: Awesome!

flovogt (approved):

matz3 (approved):

PR #628, opened 2023-07-06
[FIX] Allow usage of after/before task assignment for all standard tasks

flovogt:
This change allows custom tasks to subscribe to any standard task, even disabled ones, ensuring custom tasks' execution and the correct order.

Background information:
When transitioning a project from V2 to V3, we often need to adjust the before/after task assignments for custom tasks due to the removal of the `uglify` task. The new `minify` task now runs earlier than the previous `uglify` task.
Developers typically used `uglify` as the last enabled standard task to subscribe to. 

You can find updated documentation here: https://github.com/SAP/ui5-tooling/pull/874

JIRA: CPOUI5FOUNDATION-724

### DIFF
--- a/lib/build/definitions/application.js
+++ b/lib/build/definitions/application.js
@@ -103,7 +103,7 @@ export default function({project, taskUtil, getTask}) {
 	} else {
 		// No bundles defined. Just set task so that it can be referenced by custom tasks
 		tasks.set("generateBundle", {
-			taskFunction: async () => {}
+			taskFunction: null
 		});
 	}
 
--- a/lib/build/definitions/application.js
+++ b/lib/build/definitions/application.js
@@ -100,6 +100,11 @@ export default function({project, taskUtil, getTask}) {
 				}, Promise.resolve());
 			}
 		});
+	} else {
+		// No bundles defined. Just set task so that it can be referenced by custom tasks
+		tasks.set("generateBundle", {
+			taskFunction: async () => {}
+		});
 	}
 
 	tasks.set("generateVersionInfo", {
--- a/test/lib/build/definitions/library.js
+++ b/test/lib/build/definitions/library.js
@@ -631,7 +631,7 @@ test("buildThemes: CSS Variables enabled", (t) => {
 		"taskUtil#getBuildOption got called with correct argument");
 });
 
-test("Standard build: mocked functions for skipped tasks", async (t) => {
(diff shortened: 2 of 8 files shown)


lib/build/definitions/application.js
```diff
@@ -100,6 +100,11 @@ export default function({project, taskUtil, getTask}) {
 				}, Promise.resolve());
 			}
 		});
+	} else {
+		// No bundles defined. Just set task so that it can be referenced by custom tasks
+		tasks.set("generateBundle", {
```
d3xter666, line 105: There's also a corresponding task in `lib/build/definitions/library.js`. 
Should we also do the same there?

d3xter666 (commented):
I got it from the description that we currently face issues only with `generateBundle`, but I also can see potential issues of the same kind in the following places:

- `lib/build/definitions/library.js`
  - `generateComponentPreload` task
  - `generateBundle` task (already addressed in the comments)
  - `generateThemeDesignerResources` task
- `lib/build/definitions/themeLibrary.js`
  - `generateThemeDesignerResources` task

I'm unsure whether we need to address those now, but they seem the same.

matz3 (dismissed):
I missed some detail when recommending to just use `null` as taskFunction.
The task logger used in `runTasks` must only receive the list of task which are actually executed and logged. Otherwise it will wait forever and the build never succeeds (you should be able to verify that e.g. with openui5-sample-app).

So the `allTasks` list must not include tasks which should not be executed. So the initial solution should be fine (using `composeTaskList`) as long as the logic is added after the include/excludes are handled because such a no-op task should never be enabled.

test/lib/build/definitions/library.js
```diff
@@ -606,3 +630,89 @@ test("buildThemes: CSS Variables enabled", (t) => {
 	t.is(taskUtil.getBuildOption.getCall(0).args[0], "cssVariables",
 		"taskUtil#getBuildOption got called with correct argument");
 });
+
+test("Standard build: nulled taskFunction to skip tasks", (t) => {
+	const {project, taskUtil, getTask} = t.context;
```
flovogt, line 635: should we have a test to check assigning a custom task to any possible task?

d3xter666: Thanks! I have updated the TasksRunner's tests with that.

matz3 (approved):

PR #631, opened 2023-07-12
[FIX] Application: Fallback to manifest.appdescr_variant if manifest.json is not found

RandomByte:
This fixes a regression presumably present since ui5-project v3.0.0,
which prevented proper fallback to manifest.appdescr_variant in case no
manifest.json is found in the project.

### DIFF
--- a/lib/specifications/types/Application.js
+++ b/lib/specifications/types/Application.js
@@ -124,7 +124,7 @@ class Application extends ComponentProject {
 		try {
 			return await this._getNamespaceFromManifestJson();
 		} catch (manifestJsonError) {
-			if (manifestJsonError.cause?.code !== "ENOENT") {
+			if (manifestJsonError.code !== "ENOENT") {
 				throw manifestJsonError;
 			}
 			// No manifest.json present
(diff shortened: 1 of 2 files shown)


lib/specifications/types/Application.js
```diff
@@ -124,15 +124,15 @@ class Application extends ComponentProject {
 		try {
 			return await this._getNamespaceFromManifestJson();
 		} catch (manifestJsonError) {
-			if (manifestJsonError.code !== "ENOENT") {
+			if (manifestJsonError.cause?.code !== "ENOENT") {
```
codeworrior, line 127: In general, I don't like if the caller needs to know the potential structure of an exception chain. However, as this is inside the same module, it's  acceptable. 

Alternatively, the catch clause starting in line 224 could check for the ENOENT and throw the `err` "as is", without wrapping it. Both error messages anyhow have the same content (no added value).

For the caller this then would result in a direct check on the caught manifestJsonError.

RandomByte: Good idea, somehow I missed that 👍

codeworrior (approved):
LGTM

matz3 (commented):
Am I missing something, or is there still no test case that verifies the described fallback behaviour?

codeworrior:
@matz3 Something like [test/lib/specifications/types/Application.js#L488](https://github.com/SAP/ui5-project/blob/main/test/lib/specifications/types/Application.js#L488) ?

RandomByte:
> Alternatively, the result of `_getManifest` could be tested separately for throwing an `error` with a property `code === "NOENT"`.

This is exactly what I did: https://github.com/SAP/ui5-project/pull/631/files#diff-59c51aa221f82a38efbda063367f94782ab12d1b9d63d9e94e3917ca43f89408R589

This assertion would have failed before the fix.

tobiasqueck:
Works like a charm, thank you!

PR #635, opened 2023-07-26
[FEATURE] Allow to configure location of UI5 home directory

d3xter666:
JIRA: CPOUI5FOUNDATION-703

### DIFF
--- a/lib/config/Configuration.js
+++ b/lib/config/Configuration.js
@@ -89,12 +89,6 @@ class Configuration {
 			}
 		}
 
-		// ENV var should take precedence over the dataDir from the configuration.
-		const ui5DataDir = process.env.UI5_DATA_DIR;
-		if (ui5DataDir) {
-			config.ui5DataDir = path.resolve(path.join(ui5DataDir, ".ui5"));
-		}
-
 		return new Configuration(config);
 	}
 
--- a/lib/config/Configuration.js
+++ b/lib/config/Configuration.js
@@ -35,7 +35,7 @@ class Configuration {
 	}
 
 	/**
-	 * Configurable directory where the framework artefacts would be stored.
+	 * Configurable directory where the framework artefacts are stored.
 	 *
 	 * @public
 	 * @returns {string}
--- a/lib/config/Configuration.js
+++ b/lib/config/Configuration.js
@@ -35,7 +35,7 @@ class Configuration {
 	}
 
 	/**
-	 * Configurable directory where the artifacts would be stored.
+	 * Configurable directory where the framework artefacts would be stored.
 	 *
 	 * @public
 	 * @returns {string}
--- a/lib/config/Configuration.js
+++ b/lib/config/Configuration.js
@@ -11,13 +11,16 @@ import os from "node:os";
  */
 class Configuration {
 	#mavenSnapshotEndpointUrl;
+	#ui5DataDir;
 
 	/**
 	 * @param {object} configuration
 	 * @param {string} [configuration.mavenSnapshotEndpointUrl]
+	 * @param {string} [configuration.ui5DataDir]
 	 */
-	constructor({mavenSnapshotEndpointUrl}) {
+	constructor({mavenSnapshotEndpointUrl, ui5DataDir}) {
 		this.#mavenSnapshotEndpointUrl = mavenSnapshotEndpointUrl;
+		this.#ui5DataDir = ui5DataDir;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


flovogt (changes requested):

lib/config/Configuration.js
```diff
@@ -31,13 +34,25 @@ class Configuration {
 		return this.#mavenSnapshotEndpointUrl;
 	}
 
+	/**
+	 * Maven Repository Snapshot URL.
```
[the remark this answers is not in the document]
d3xter666: Thanks!

test/lib/config/Configuration.js
```diff
@@ -28,13 +28,14 @@ test.afterEach.always((t) => {
 	esmock.purge(t.context.Configuration);
 });
 
-test.serial("Build configuration with defaults", (t) => {
+test("Build configuration with defaults", (t) => {
```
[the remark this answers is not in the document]
d3xter666: My mistake. I was debugging something and forgot to restore it. Thanks!

lib/graph/helpers/ui5Framework.js
```diff
@@ -363,13 +364,16 @@ export default {
...
+
 		// Note: version might be undefined here and the Resolver will throw an error when calling
 		// #install and it can't be resolved via the provided library metadata
 		const resolver = new Resolver({
 			cwd: rootProject.getRootPath(),
 			version,
 			providedLibraryMetadata,
-			cacheMode
+			cacheMode,
+			ui5HomeDir: config.getUi5DataDir()
```
flovogt, line 376: so now, we have the config `ui5-data-dir` but we map it directly to `ui5-home-dir`. Is this intended?

lib/config/Configuration.js
```diff
@@ -31,13 +34,24 @@ class Configuration {
 		return this.#mavenSnapshotEndpointUrl;
 	}
 
+	/**
+	 * Configurable directory where the artifacts would be stored.
```
flovogt, line 38: what do we define as "artefacts"? these are the framework libs isn'´t?

d3xter666: By artifacts, I mean everything. It's basically the `.ui5` directory whith the following structure inside:
```yaml
- framework
   - artifacts
   - locks
   - metadata
   - packages
   - staging
```

Maybe, a better wording should be chosen? 🤔

flovogt: i would propose "framework artefacts"

lib/config/Configuration.js
```diff
@@ -11,13 +11,16 @@ import os from "node:os";
...
 	#mavenSnapshotEndpointUrl;
+	#ui5DataDir;
 
 	/**
 	 * @param {object} configuration
 	 * @param {string} [configuration.mavenSnapshotEndpointUrl]
+	 * @param {string} [configuration.ui5DataDir]
 	 */
-	constructor({mavenSnapshotEndpointUrl}) {
+	constructor({mavenSnapshotEndpointUrl, ui5DataDir}) {
```
flovogt, line 21: currently the configuration does not offer a default value but in real the home dir has a default value.... should we not align this?
so moving the defaulting also to config here?

lib/config/Configuration.js
d3xter666: Thanks!

lib/config/Configuration.js
d3xter666: Thanks for bringin it up!
This would mean that the `.ui5rc` file would always receive a value for the `ui5DataDir` and this setting would always be in there, regardless of the fact it's set intentionally or not.
Do we want to always have this in the `.ui5rc` file?

Also, the resolvers (i.e. `Openui5Resolver.js`) that utilize this directory are exposed through the `package.json` and in the best case we would need to double maintain the `ui5HomeDir`- in the `AbstractResolver.js` and in the `Configuration.js`. 
We cannot delete the default value from the `AbstractResolver.js (-> Openui5Resolver.js)` where it's populated currently, because that would be an incompatible change.

What do you think? Which way to choose? :D

flovogt: @RandomByte any opinion on the preferred way?

RandomByte: Good question. I think we discussed this as part of [RFC 0013 Configuration](https://github.com/SAP/ui5-tooling/blob/main/rfcs/0013-configuration.md#configuration-module)

There it says:
> [...] **Parameters shall not have default values.** [...]

So I think we should not add a defaulting here.

This means that Configuration.js is not the single source of truth for which ui5DataDir is used, but that's our compromise. Every consumer needs to apply the defaulting themselves, or we add a dedicated class/function somewhere to centralize that.

Actually, the environment variable must also not be read and applied in Configuration.js but on the consumer side. As per the same RFC:
> There shall **not be any fallback mechanism to environment variables** or similar inside the Configuration.js module. [...]

d3xter666: Ah, thanks a lot for clearing this out!

lib/config/Configuration.js
```diff
@@ -31,13 +34,24 @@ class Configuration {
 		return this.#mavenSnapshotEndpointUrl;
 	}
 
+	/**
+	 * Configurable directory where the framework artefacts would be stored.
```
flovogt, line 38:
```suggestion
	 * Configurable directory where the framework artefacts are stored.
```

flovogt (changes requested):

flovogt (commented):
LGTM, waiting for review by @matz3 @RandomByte

lib/graph/helpers/ui5Framework.js
```diff
@@ -363,13 +365,20 @@ export default {
 			});
 		}
 
+		const config = await Configuration.fromFile();
+		// ENV var should take precedence over the dataDir from the configuration.
+		const ui5HomeDir = process.env.UI5_DATA_DIR ?
+			path.resolve(path.join(process.env.UI5_DATA_DIR, ".ui5")) :
```
RandomByte, line 371: I would expect that we use `UI5_DATA_DIR` as the "home dir" and don't create a `.ui5` directory inside of it.

RandomByte (changes requested):

lib/graph/helpers/ui5Framework.js
d3xter666: Done

RandomByte (approved):

PR #642, opened 2023-08-16
[FIX] Resolve UI5 data directory relative to project

matz3:
The UI5 data directory (previously called UI5 home directory) should
always be resolved relative to the project directory, which is where
the package.json file is located at.

This change also adds a list of options to the Configuration class to be
used by the CLI config command to allow all available config options to
be set or retrieved.

Follow-up of https://github.com/SAP/ui5-project/pull/635

JIRA: CPOUI5FOUNDATION-715
JIRA: CPOUI5FOUNDATION-716

### DIFF
--- a/test/lib/graph/helpers/ui5Framework.integration.js
+++ b/test/lib/graph/helpers/ui5Framework.integration.js
@@ -125,11 +125,19 @@ test.beforeEach(async (t) => {
 		"../../../../lib/specifications/Specification.js": t.context.Specification
 	});
 
+	// Stub os homedir to prevent that the actual ~/.ui5rc from being used in tests
+	t.context.Configuration = await esmock.p("../../../../lib/config/Configuration.js", {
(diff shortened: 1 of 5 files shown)


test/lib/graph/helpers/ui5Framework.integration.js
```diff
@@ -125,11 +125,19 @@ test.beforeEach(async (t) => {
 		"../../../../lib/specifications/Specification.js": t.context.Specification
 	});
 
+	// Stub os homedir to prevent that the actual ~/.ui5rc from being used in tests
```
RandomByte, line 128:
```suggestion
	// Stub os homedir to prevent the actual ~/.ui5rc from being used in tests
```

RandomByte (changes requested):
LGTM, just a minor proposal

RandomByte (approved):

PR #649, opened 2023-08-24
[FEATURE] Resolvers: Allow ranges / npm tags for version resolution

RandomByte:
The framework version resolution now supports all valid node semver
ranges and uses tags from the corresponding registry for version
resolution (not supported for Sapui5MavenSnapshotResolver).

JIRA: CPOUI5FOUNDATION-722

### DIFF
--- a/lib/ui5Framework/AbstractResolver.js
+++ b/lib/ui5Framework/AbstractResolver.js
@@ -258,6 +258,7 @@ class AbstractResolver {
 			}
 		}
 
+		// Covers versions and ranges, as versions are also valid ranges
 		if (semver.validRange(version)) {
 			return version;
 		}
--- a/lib/ui5Framework/AbstractResolver.js
+++ b/lib/ui5Framework/AbstractResolver.js
@@ -215,35 +215,54 @@ class AbstractResolver {
 	}
 
 	static async resolveVersion(version, {ui5HomeDir, cwd} = {}) {
-		if (version === "latest") {
-			const tagVersion = this.fetchTag("latest", {ui5HomeDir, cwd});
-			if (tagVersion) {
-				return tagVersion;
+		// Don't allow nullish values
+		// An empty string is a valid semver range that converts to "*", which should not be supported
+		if (!version) {
+			throw new Error(`Framework version specifier "${version}" is incorrect or not supported`);
+		}
+
+		const isSnapshotVersion = version.toLowerCase().endsWith("-snapshot");
+		if (isSnapshotVersion) {
+			const versionMatch = version.match(VERSION_RANGE_REGEXP);
+			if (versionMatch) {
+				// For snapshot version ranges we need to insert a stand-in "x" for the patch level
+				// and - in case none is provided - another "x" for the major version in order to
+				// convert it to a valid semver range:
+				// "1-SNAPSHOT" becomes "1.x.x-SNAPSHOT" and "1.112-SNAPSHOT" becomes "1.112.x-SNAPSHOT"
+				version = `${versionMatch[1]}.${versionMatch[2] || "x"}.x-SNAPSHOT`;
 			}
 		}
 
 		let spec;
-		const isSnapshotVersion = version.toLowerCase().endsWith("-snapshot");
-		if (version === "latest" || version === "latest-snapshot") {
-			// Use a wildcard to resolve to the latest available version
-			spec = "*";
-		} else if (SEMVER_VERSION_REGEXP.test(version)) {
-			// Fully qualified version, can be used directly
+
+		if (semver.validRange(version)) {
+			// Note: Fully qualified versions are also valid ranges
 			spec = version;
-		} else {
-			const versionMatch = version.match(VERSION_RANGE_REGEXP);
-			if (versionMatch) {
-				if (isSnapshotVersion) {
-					// For snapshot version ranges we need to insert a stand-in "x" for the patch level
-					// and - in case none is provided - another "x" for the major version in order to
-					// make the semver check work: "1-SNAPSHOT" or "1.112-SNAPSHOT" becomes "1.112.x-SNAPSHOT"
-					spec = `${versionMatch[1]}.${versionMatch[2] || "x"}.x-SNAPSHOT`;
-				} else {
-					spec = version;
+		} else if (encodeURIComponent(version) === version) {
+			// Valid tag name (same check as npm does)
+			const allTags = await this.fetchAllTags({ui5HomeDir, cwd});
+			if (!allTags) {
+				// Resolver doesn't support tags (e.g. Sapui5MavenSnapshotResolver)
+				if (version === "latest" || version === "latest-snapshot") {
+					// Only latest and latest-snapshot are supported which both resolve
+					// to the latest available version.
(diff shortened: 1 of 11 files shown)


lib/ui5Framework/AbstractResolver.js
```diff
@@ -258,9 +242,60 @@ class AbstractResolver {
...
+		if (semver.validRange(version)) {
+			return version;
+		}
+
+		// Check for invalid tag name (same check as npm does)
+		if (encodeURIComponent(version) !== version) {
+			return null;
+		}
+
+		const allTags = await this.fetchAllTags({ui5HomeDir, cwd});
```
d3xter666, line 270:
```suggestion
		const allTags = await AbstractResolver.fetchAllTags({ui5HomeDir, cwd});
```

lib/ui5Framework/AbstractResolver.js
```diff
@@ -258,9 +242,60 @@ class AbstractResolver {
 				`Could not resolve framework version ${version}. ` +
 				`Make sure the version is valid and available in the configured registry.`);
 		}
+
 		return resolvedVersion;
 	}
 
+	static async _getVersionSpec(version, {ui5HomeDir, cwd}) {
+		if (this._isSnapshotVersionOrRange(version)) {
```
d3xter666, line 250:
```suggestion
		if (AbstractResolver._isSnapshotVersionOrRange(version)) {
```

lib/ui5Framework/AbstractResolver.js
```diff
@@ -215,33 +206,26 @@ class AbstractResolver {
...
-			} else {
-				throw new Error(`Framework version specifier "${version}" is incorrect or not supported`);
-			}
+		// Don't allow nullish values
+		// An empty string is a valid semver range that converts to "*", which should not be supported
+		if (!version) {
+			throw new Error(`Framework version specifier "${version}" is incorrect or not supported`);
+		}
+
+		const spec = await this._getVersionSpec(version, {ui5HomeDir, cwd});
```
d3xter666, line 215:
```suggestion
		const spec = await AbstractResolver._getVersionSpec(version, {ui5HomeDir, cwd});
```

lib/ui5Framework/AbstractResolver.js
```diff
@@ -258,9 +242,60 @@ class AbstractResolver {
...
+
+		const allTags = await this.fetchAllTags({ui5HomeDir, cwd});
+
+		if (!allTags) {
+			// Resolver doesn't support tags (e.g. Sapui5MavenSnapshotResolver)
+			// Only latest and latest-snapshot are supported which both resolve
+			// to the latest available version.
+			// See "isSnapshotVersionOrRange" for -snapshot handling
+			if ((version === "latest" || version === "latest-snapshot")) {
+				return "*";
```
d3xter666, line 278: The asterisk (*) would resolve `latest`, but would this be correct for `latest-snapshot`?
What about `x.x.x-SNAPSHOT` for `latest-snapshot`?

lib/ui5Framework/AbstractResolver.js
```diff
@@ -215,33 +206,26 @@ class AbstractResolver {
...
+
+		const spec = await this._getVersionSpec(version, {ui5HomeDir, cwd});
+
+		// For all invalid cases which are not explicitly handled in _getVersionSpec
+		if (!spec) {
+			throw new Error(`Framework version specifier "${version}" is incorrect or not supported`);
 		}
+
 		const versions = await this.fetchAllVersions({ui5HomeDir, cwd});
 		const resolvedVersion = semver.maxSatisfying(versions, spec, {
```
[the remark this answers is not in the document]
matz3: Semver has the concept of pre-releases. And the -SNAPSHOT suffix is such a pre-release identifier.
So I'm not sure why you mean with "does not understand".

Yes, for the snapshot resolver we will only have versions with the SNAPSHOT pre-release identifier.

lib/ui5Framework/AbstractResolver.js
matz3: This is the existing behaviour. Using an asterisk and "includePrerelease" will resolve to the highest version.

`x.x.x-SNAPSHOT` sounds like a better fit, but it actually is handled the same by semver (`require("semver").validRange("x.x.x-SNAPSHOT") === "*"`).
Therefore, if there would be a higher released version (no -SNAPSHOT), it would always be picked, even with that range.

Usually that is not the case as the highest version always has a snapshot version available. But even if this wouldn't be the case, our snapshot resolver filters for -SNAPSHOT versions so it will always resolve to the highest snapshot.

test/lib/ui5framework/AbstractResolver.js
```diff
@@ -728,17 +726,19 @@ test.serial("AbstractResolver: Static resolveVersion resolves 'latest-snapshot'"
...
 });
 
-test.serial("AbstractResolver: Static resolveVersion resolves 'MAJOR.MINOR-SNAPSHOT'", async (t) => {
+test.serial("AbstractResolver: Static resolveVersion includes non-prereleases for 'latest-snapshot'", async (t) => {
+	// Realistically this should never happen, since the Sapui5MavenSnapshotResolver would never return
+	// non-snapshot versions. This test therefore simply illustrates the current behavior for this theoretic case
 	const {MyResolver} = t.context;
 	const fetchAllVersionsStub = sinon.stub(MyResolver, "fetchAllVersions")
-		.returns(["1.75.0-SNAPSHOT", "1.75.1-SNAPSHOT", "1.76.0-SNAPSHOT", "1.76.1-SNAPSHOT"]);
+		.returns(["1.76.0", "1.77.0", "1.78.0", "1.79.0-SNAPSHOT", "1.79.1"]);
```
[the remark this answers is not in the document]
matz3: In most Maven registries the -SNAPSHOT versions are not deleted right after a higher minor/patch is released and also not after a new -SNAPSHOT version for a higher minor/patch is deployed. SNAPSHOT versions remain in the registry for some time and will be deleted when they are not updated for some amount of time (I think a few weeks).

test/lib/ui5framework/AbstractResolver.js
```diff
@@ -728,17 +726,19 @@ test.serial("AbstractResolver: Static resolveVersion resolves 'latest-snapshot'"
...
 
-	const version = await MyResolver.resolveVersion("1.75-SNAPSHOT", {
+	const version = await MyResolver.resolveVersion("latest-snapshot", {
 		cwd: "/cwd",
 		ui5HomeDir: "/ui5HomeDir"
 	});
 
-	t.is(version, "1.75.1-SNAPSHOT", "Resolved version should be correct");
+	t.is(version, "1.79.1", "Resolved version should be correct");
```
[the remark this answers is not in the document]
matz3: It means latest -SNAPSHOT version. And this behaviour is guaranteed as for a version/range/tag ending with "-SNAPSHOT" the snapshot resolver is picked, which will only provide SNAPSHOT versions. That's what the comment above states.

Note that this test already existed before (which is not easily visible from the diff) and the goal of this PR is to keep any existing behaviour as is.

d3xter666 (approved):

PR #652, opened 2023-09-01
[FIX] Ensure usage of provided UI5 data dir

matz3:
Also moves check for framework libraries to earlier point as further
steps such as choosing a resolver and calling resolveVersion are only
needed when libraries are referenced.
This aligns the behavior with not having a framework configuration at
all but still passing "versionOverride" which is then also not
resolved/checked for correctness.

### DIFF
--- a/lib/graph/helpers/ui5Framework.js
+++ b/lib/graph/helpers/ui5Framework.js
@@ -294,6 +294,7 @@ export default {
 		const rootProject = projectGraph.getRoot();
 		const frameworkName = rootProject.getFrameworkName();
 		const frameworkVersion = rootProject.getFrameworkVersion();
+		const cwd = rootProject.getRootPath();
 
 		// It is allowed to omit the framework version in ui5.yaml and only provide one via the override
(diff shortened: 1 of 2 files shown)


lib/graph/helpers/ui5Framework.js
```diff
@@ -339,20 +348,26 @@ export default {
 			Resolver = (await import("../../ui5Framework/Sapui5Resolver.js")).default;
 		}
 
+		// ENV var should take precedence over the dataDir from the configuration.
+		let ui5DataDir = process.env.UI5_DATA_DIR;
+		if (!ui5DataDir) {
+			const config = await Configuration.fromFile();
+			ui5DataDir = config.getUi5DataDir();
+		}
+		if (ui5DataDir) {
```
flovogt, line 357: what happens if ui5DataDir is nullish?

matz3: Then that is passed to the API and should be handed the same as if it's undefined.

flovogt (approved):

PR #677, opened 2023-11-10
[FEATURE] TaskUtil: Add 'force' flag to cleanup task callback

d3xter666:
JIRA: CPOUI5FOUNDATION-751
\@ui5/builder PR that depends on this change: https://github.com/SAP/ui5-builder/pull/953

When cleaning up after tasks run, it can happen in couple cases:
- When the project is build
- When there's a SIGTERM signal
- (Maybe) on some error

We need to be able to distinguish between those cases as in some places we rely on graceful termination and cleanup. Especially, after the project has been build successfully. On the other hand when the user presses `Ctrl + C`, for example, we need to enforce the termination.

This cahnge provides this context.

### DIFF
--- a/lib/build/ProjectBuilder.js
+++ b/lib/build/ProjectBuilder.js
@@ -390,7 +390,8 @@ class ProjectBuilder {
 
 	async _executeCleanupTasks(force) {
 		this.#log.info("Executing cleanup tasks...");
-		await this._buildContext.executeCleanupTasks(forceCleanup);
+
+		await this._buildContext.executeCleanupTasks(force);
 	}
 
 	_registerCleanupSigHooks() {
--- a/lib/build/ProjectBuilder.js
+++ b/lib/build/ProjectBuilder.js
@@ -388,9 +388,9 @@ class ProjectBuilder {
 		}));
 	}
 
-	async _executeCleanupTasks() {
+	async _executeCleanupTasks(exitCode) {
 		this.#log.info("Executing cleanup tasks...");
-		await this._buildContext.executeCleanupTasks();
+		await this._buildContext.executeCleanupTasks(exitCode);
 	}
 
 	_registerCleanupSigHooks() {
@@ -398,7 +398,7 @@ class ProjectBuilder {
 		function createListener(exitCode) {
 			return function() {
 				// Asynchronously cleanup resources, then exit
-				that._executeCleanupTasks().then(() => {
+				that._executeCleanupTasks({exitCode}).then(() => {
 					process.exit(exitCode);
 				});
 			};
--- a/lib/build/ProjectBuilder.js
+++ b/lib/build/ProjectBuilder.js
@@ -388,9 +388,9 @@ class ProjectBuilder {
 		}));
 	}
 
-	async _executeCleanupTasks(exitCode) {
+	async _executeCleanupTasks(options) {
 		this.#log.info("Executing cleanup tasks...");
-		await this._buildContext.executeCleanupTasks(exitCode);
+		await this._buildContext.executeCleanupTasks(options);
 	}
 
 	_registerCleanupSigHooks() {
--- a/lib/build/ProjectBuilder.js
+++ b/lib/build/ProjectBuilder.js
@@ -388,9 +388,9 @@ class ProjectBuilder {
 		}));
 	}
 
-	async _executeCleanupTasks(options) {
+	async _executeCleanupTasks(forceCleanup) {
 		this.#log.info("Executing cleanup tasks...");
-		await this._buildContext.executeCleanupTasks(options);
+		await this._buildContext.executeCleanupTasks(forceCleanup);
 	}
 
 	_registerCleanupSigHooks() {
@@ -398,7 +398,7 @@ class ProjectBuilder {
 		function createListener(exitCode) {
 			return function() {
 				// Asynchronously cleanup resources, then exit
-				that._executeCleanupTasks({exitCode}).then(() => {
+				that._executeCleanupTasks(true).then(() => {
 					process.exit(exitCode);
 				});
 			};
--- a/lib/build/ProjectBuilder.js
+++ b/lib/build/ProjectBuilder.js
@@ -388,7 +388,7 @@ class ProjectBuilder {
 		}));
 	}
 
-	async _executeCleanupTasks(forceCleanup) {
+	async _executeCleanupTasks(force) {
 		this.#log.info("Executing cleanup tasks...");
 		await this._buildContext.executeCleanupTasks(forceCleanup);
 	}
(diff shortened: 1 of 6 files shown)


lib/build/ProjectBuilder.js
```diff
@@ -388,17 +388,17 @@ class ProjectBuilder {
...
+		await this._buildContext.executeCleanupTasks(options);
 	}
 
 	_registerCleanupSigHooks() {
 		const that = this;
 		function createListener(exitCode) {
 			return function() {
 				// Asynchronously cleanup resources, then exit
-				that._executeCleanupTasks().then(() => {
+				that._executeCleanupTasks({exitCode}).then(() => {
```
RandomByte, line 401: Any reason the task might need the exact exit code? I'm wondering whether we should rather have a `force` boolean to make the intention more clear. Note that this is [public API](https://sap.github.io/ui5-tooling/stable/api/@ui5_project_build_helpers_TaskUtil.html#registerCleanupTask) provided to custom tasks, so it should be intuitive.

d3xter666: Agree with that. I was thinking that we wanted to keep the API more open for changes, but a `force` param seems logical and correct.

lib/build/ProjectBuilder.js
```diff
@@ -388,17 +388,17 @@ class ProjectBuilder {
 		}));
 	}
 
-	async _executeCleanupTasks() {
+	async _executeCleanupTasks(forceCleanup) {
```
RandomByte, line 391: Hm, here we call it `forceCleanup` and in https://github.com/SAP/ui5-builder/pull/953/commits/2f8181f93d86d7603d853e8f7e625ec58d7d6aa4#diff-58ad88f33b04bd528f373d803ff59e814795eaff5cb0fee7f90cb2732751d294R31 it's `forceTerminate`

Should we just settle with `force`?

```suggestion
	async _executeCleanupTasks(force) {
```

test/lib/build/ProjectBuilder.js
```diff
@@ -727,7 +727,7 @@ test("_executeCleanupTasks", async (t) => {
 	const executeCleanupTasksStub = sinon.stub(builder._buildContext, "executeCleanupTasks");
 	await builder._executeCleanupTasks();
 	t.is(executeCleanupTasksStub.callCount, 1, "BuildContext#executeCleanupTasks got called once");
-	t.deepEqual(executeCleanupTasksStub.getCall(0).args, [],
+	t.deepEqual(executeCleanupTasksStub.getCall(0).args, [undefined],
```
RandomByte, line 730: LGTM, but could you please add another test where we call `_executeCleanupTasks` with `force=true` and test that the callback correctly receives the flag?

RandomByte (changes requested):

test/lib/build/ProjectBuilder.js
d3xter666: Done

RandomByte (dismissed):

RandomByte:
Maybe we should have this as a `[FEATURE]` in the changelog?

RandomByte:
We should also update the JSDoc in TaskUtil: https://github.com/SAP/ui5-project/blob/main/lib/build/helpers/TaskUtil.js#L175

d3xter666:
> We should also update the JSDoc in TaskUtil: https://github.com/SAP/ui5-project/blob/main/lib/build/helpers/TaskUtil.js#L175

I have added JSDoc for the TaskUtil's cleanup callback.

@KlattG would you take a look, please?

RandomByte (approved):

lib/build/helpers/TaskUtil.js
```diff
@@ -163,6 +163,16 @@ class TaskUtil {
 		return this._projectBuildContext.getOption(key);
 	}
 
+	/**
+	 * Callback that will be executed once the build is finished
+	 *
+	 * @public
+	 * @callback @ui5/project/build/helpers/TaskUtil~cleanupTaskCallback
+	 * @param {boolean} force Specifies whether the cleanup callback should
```
matz3, line 171: The current implementation doesn't include a default behaviour (to false), so as `executeCleanupTasks` is called with undefined, the callbacks also receive undefined, but this does not fit to the type `{boolean}` in here.

matz3 (changes requested):

lib/build/helpers/BuildContext.js
```diff
@@ -80,9 +80,9 @@ class BuildContext {
 		return projectBuildContext;
 	}
 
-	async executeCleanupTasks() {
+	async executeCleanupTasks(force) {
 		await Promise.all(this._projectBuildContexts.map((ctx) => {
-			return ctx.executeCleanupTasks();
+			return ctx.executeCleanupTasks(force);
 		}));
```
matz3, line 86: I would suggest to default to `false` in here, so that the callbacks are always called with a boolean.

RandomByte: Done

lib/build/helpers/TaskUtil.js
```diff
@@ -163,6 +163,16 @@ class TaskUtil {
 		return this._projectBuildContext.getOption(key);
 	}
 
+	/**
+	 * Callback that will be executed once the build is finished
```
KlattG, line 167:
```suggestion
	 * Callback that is executed once the build has finished
```

lib/build/helpers/TaskUtil.js
```diff
@@ -172,7 +182,8 @@ class TaskUtil {
 	 * This method is only available to custom task extensions defining
 	 * <b>Specification Version 2.2 and above</b>.
 	 *
-	 * @param {Function} callback Callback to register. If it returns a Promise, it will be waited for
+	 * @param {@ui5/project/build/helpers/TaskUtil~cleanupTaskCallback} callback Callback to register.
+	 * 			If it returns a Promise, it will be waited for
```
KlattG, line 186:
```suggestion
	 * @param {@ui5/project/build/helpers/TaskUtil~cleanupTaskCallback} callback Callback to register; it will be waited for if it returns a Promise
	 * 			
```

lib/build/helpers/TaskUtil.js
```diff
@@ -163,6 +163,16 @@ class TaskUtil {
...
 	}
 
+	/**
+	 * Callback that will be executed once the build is finished
+	 *
+	 * @public
+	 * @callback @ui5/project/build/helpers/TaskUtil~cleanupTaskCallback
+	 * @param {boolean} force Specifies whether the cleanup callback should
+	 * 							gracefully wait for certain jobs to be completed (<code>false</code>)
+	 * 							or enforce immediate termination (<code>true</code>).
```
KlattG, line 173:
```suggestion
	 * @param {boolean} force Whether the cleanup callback should
	 * 							gracefully wait for certain jobs to be completed (<code>false</code>)
	 * 							or enforce immediate termination (<code>true</code>)
```

KlattG (changes requested):

matz3 (approved):

RandomByte (approved):

PR #706, opened 2024-02-23
[BREAKING] Set default workspaceName to "default" for API usage

d3xter666:
BREAKING CHANGE:
Set default workspaceName to "default" for API usage (https://github.com/SAP/ui5-project/pull/586)

JIRA: CPOUI5FOUNDATION-802
Relates to: https://github.com/SAP/ui5-tooling/issues/701

### DIFF
--- a/lib/graph/graph.js
+++ b/lib/graph/graph.js
@@ -54,12 +54,15 @@ export async function graphFromPackageDependencies({
 	cwd = cwd ? path.resolve(cwd) : process.cwd();
 	rootConfigPath = utils.resolveConfigPath(cwd, rootConfigPath);
 
-	const workspace = await createWorkspace({
-		cwd,
-		name: workspaceName,
-		configObject: workspaceConfiguration,
-		configPath: workspaceConfigPath,
-	});
+	let workspace;
+	if (workspaceName || workspaceConfiguration) {
+		workspace = await createWorkspace({
+			cwd,
+			name: workspaceName,
+			configObject: workspaceConfiguration,
+			configPath: workspaceConfigPath
+		});
+	}
 
 	const provider = new NpmProvider({
 		cwd,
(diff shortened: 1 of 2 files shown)


lib/graph/graph.js
```diff
@@ -54,15 +54,12 @@ export async function graphFromPackageDependencies({
 	cwd = cwd ? path.resolve(cwd) : process.cwd();
 	rootConfigPath = utils.resolveConfigPath(cwd, rootConfigPath);
 
-	let workspace;
-	if (workspaceName || workspaceConfiguration) {
```
matz3, line 58: `workspaceName` can still be falsy, so I think the if-clause shouldn't be removed.

We should actually also allow `null` as a value. See my code suggestion above.

lib/graph/graph.js
```diff
@@ -29,8 +29,8 @@ const log = getLogger("generateProjectGraph");
  * @param {string} [options.versionOverride] Framework version to use instead of the one defined in the root project
  * @param {string} [options.resolveFrameworkDependencies=true]
  * 		Whether framework dependencies should be added to the graph
- * @param {string} [options.workspaceName]
- * 		Name of the workspace configuration that should be used if any is provided
+ * @param {string} [options.workspaceName=default]
```
matz3, line 32: We should also make clear that with `null` the usage of the default workspace can be skipped. The CLI currently already passes `null` if `--no-workspace` option is set.

```suggestion
 * @param {string|null} [options.workspaceName=default]
```

lib/graph/graph.js
d3xter666: Thanks! I have updated the tests to reflect that

lib/graph/graph.js
```diff
@@ -43,7 +43,7 @@ const log = getLogger("generateProjectGraph");
 export async function graphFromPackageDependencies({
 	cwd, rootConfiguration, rootConfigPath,
 	versionOverride, cacheMode, resolveFrameworkDependencies = true,
-	workspaceName /* TODO 4.0: default workspaceName to "default" ? */,
+	workspaceName="default",
```
matz3, line 46: There's no test assertion that makes sure "default" is used as default value. Renaming this string does not fail any test case in `test/lib/graph/graph.js`.

matz3 (changes requested):

lib/graph/graph.js
d3xter666: Thanks! I have created a test for the "default" workspace name

RandomByte (approved):
LGTM

PR #725, opened 2024-05-21
[FEATURE] Bundling: Remove option 'usePredefineCalls'

flovogt:
Up until UI5 Tooling v3, the bundle option "usePredefineCalls" defaults to "false" and has to be activated explicitly in a custom bundle configuration.
For default bundles, such as Component-preload or self-contained bundle (sap-ui-custom.js), there is no way to use the option apart from re-defining the whole bundle via custom bundle definition.

With UI5 Tooling v4 bundles are generated with the usage sap.ui.predefine calls instead of the former default function wrapper. This leads to smaller bundle sizes and less overhead at runtime.

As the option only affects the internal handling of bundling without affecting the actual users,
the option is removed completely, instead of just changing the default value of the option.

Related to https://github.com/SAP/ui5-builder/pull/1021.

Documentation update covered in https://github.com/SAP/ui5-tooling/pull/957.

JIRA: CPOUI5FOUNDATION-760

### DIFF
--- a/lib/validation/schema/specVersion/kind/project.json
+++ b/lib/validation/schema/specVersion/kind/project.json
@@ -397,10 +397,6 @@
 					"type": "boolean",
 					"default": false
 				},
-				"usePredefineCalls": {
-					"type": "boolean",
-					"default": false
-				},
 				"numberOfParts": {
 					"type": "number",
 					"default": 1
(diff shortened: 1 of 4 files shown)


lib/validation/schema/specVersion/kind/project.json
```diff
@@ -397,10 +397,6 @@
 					"type": "boolean",
 					"default": false
 				},
-				"usePredefineCalls": {
```
matz3, line 400: This changes existing specVersions, which is not intended

matz3 (changes requested):

flovogt:
The update of the schema will be handled by a new specVersion 4.0 (JIRA: CPOUI5FOUNDATION-835). Therefore closing this PR.

PR #728, opened 2024-06-07
[BREAKING] Make '@ui5/builder' an optional peerDependency

RandomByte:
This decoupling allows the use of @ui5/project without having to install
@ui5/builder with all its transitive dependencies in cases where no
build functionality is required. An example for that is UI5 linter.

BREAKING CHANGE:
Consumers of the Node.js API that make use of the `ProjectGraph#build` method need to declare a dependency to `@ui5/builder` in their respective package.json. The package manager should ensure that the version fullfills the range specified in the package.json of `@ui5/project`.

JIRA: CPOUI5FOUNDATION-854

### DIFF
--- a/lib/graph/ProjectGraph.js
+++ b/lib/graph/ProjectGraph.js
@@ -513,7 +513,13 @@ class ProjectGraph {
 
 	async _getTaskRepository() {
 		if (!this._taskRepository) {
-			this._taskRepository = await import("@ui5/builder/internal/taskRepository");
+			try {
+				this._taskRepository = await import("@ui5/builder/internal/taskRepository");
+			} catch (err) {
+				throw new Error(
+					`Failed to load task repository. Missing dependency to '@ui5/builder'? ` +
+					`Error: ${err.message}`);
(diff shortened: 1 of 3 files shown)


d3xter666:
LGTM

RandomByte:
I tested this with UI5 linter, installing this package from a locally built tarball. As expected, `@ui5/builder` is not longer installed as well as the large transitive dependencies like terser and jsdoc.

Within ui5-project this seems to behave like a dev-dependency, in the sense that it is installed and can be used for tests.

To be checked: How does our automated update script treat this? Can it properly update the peerDependencies range in the package.json when a new version of `@ui5/builder` is installed?

matz3 (approved):

RandomByte:
> To be checked: How does our automated update script treat this? Can it properly update the peerDependencies range in the package.json when a new version of `@ui5/builder` is installed?

I checked on the update script and it should correctly update the range in peerDependencies as well (it simply executes `npm install package@version`)

PR #733, opened 2024-06-13
[FEATURE] Apply specVersion defaults from ui5.yaml.json schema

d3xter666:
JIRA: CPOUI5FOUNDATION-835


d3xter666:
Hi @KlattG,

Would you take a look at this one, please?

Cheers

lib/validation/validator.js
```diff
@@ -114,6 +137,51 @@ export async function validate(options) {
 	await _validate("ui5", options);
 }
 
+/**
+ * Validates the given ui5 configuration and returns default values if not provided.
```
KlattG, line 141:
```suggestion
 * Validates the given ui5 configuration and returns default values if none are provided.
```

lib/validation/validator.js
```diff
@@ -114,6 +137,51 @@ export async function validate(options) {
...
+ *
+ * @public
+ * @function
+ * @static
+ * @param {object} options
+ * @param {object} options.config UI5 Configuration to validate
```
KlattG, line 147:
```suggestion
 * @param {object} options.config The UI5 Configuration to validate
```

lib/validation/validator.js
```diff
@@ -114,6 +137,51 @@ export async function validate(options) {
...
+ *   Rejects with a {@link @ui5/project/validation/ValidationError ValidationError}
+ *   when the validation fails.
+ * @returns {Promise<options>} Returns a Promise that resolves when the validation succeeds
+ */
+export async function getDefaults(options) {
+	return await _validateAndSetDefaults("ui5", options);
+}
+
+/**
+ * Enhances bundleDefinition by adding missing properties with their respective default values
```
KlattG, line 164:
```suggestion
 * Enhances bundleDefinition by adding missing properties with their respective default values.
```

KlattG (changes requested):
Just some minor suggestions, otherwise LGTM

KlattG (approved):
LGTM

lib/build/definitions/library.js
```diff
@@ -120,15 +121,18 @@ export default function({project, taskUtil, getTask}) {
 			taskFunction: async ({workspace, dependencies, taskUtil, options}) => {
 				const generateBundleTask = await getTask("generateBundle");
 				return bundles.reduce(function(sequence, bundle) {
-					return sequence.then(function() {
+					return sequence.then(async function() {
+						const {bundleDefinition, bundleOptions} = await enhanceBundleWithDefaults(
```
matz3, line 125: Wouldn't it be more efficient to process all bundle configs at once before running the promise sequence?

d3xter666: Done

matz3: With that, I also meant to apply the defaults for an array of bundle configurations (as it's designed in the schema), not one by one. Or is there a reason why it has to be done one by one?

d3xter666: Ah, thanks!

I didn't see that initially! Now should be ok

lib/validation/validator.js
```diff
@@ -114,6 +137,48 @@ export async function validate(options) {
...
+ * @returns {Promise<options>} Returns a Promise that resolves when the validation succeeds
+ */
+export async function getDefaults(options) {
+	return await _validateAndSetDefaults("ui5", options);
+}
+
+/**
+ * Enhances bundleDefinition by adding missing properties with their respective default values.
+ *
+ * @param {module:@ui5/builder/processors/bundlers/moduleBundler} bundles Bundles to be enhanced
```
matz3, line 166: This type doesn't seem right. It points to the processor function.

lib/validation/validator.js
```diff
@@ -91,6 +93,27 @@ async function _validate(schemaName, options) {
...
+			return new Validator({Ajv, ajvErrors, ajvConfig: {useDefaults: true}, schemaName});
+		})();
+	}
+
+	// When AJV is configured with useDefaults: true, it may add properties to the
+	// provided configuration that were not initially present. This behavior can
+	// lead to unexpected side effects and potential issues. To avoid these
+	// problems, we create a copy of the configuration. If we need the altered
+	// configuration later, we return this copied version.
+	const optionsCopy = {...options};
```
RandomByte, line 110: I suppose we should create a deep copy? Or are we only concerned about root level changes?

RandomByte (changes requested):

lib/validation/validator.js
d3xter666: Thanks!

lib/validation/validator.js
d3xter666: Now should be the correct one

RandomByte (approved):
LGTM

PR #775, opened 2024-10-14
[INTERNAL] TaskRunner: Remove unused parameter

RandomByte:
This was noticed during the TypeScript PoC. No impact on actual runtime
behavior.

### DIFF
--- a/lib/build/TaskRunner.js
+++ b/lib/build/TaskRunner.js
@@ -438,13 +438,12 @@ class TaskRunner {
 	 * @private
 	 * @param {string} taskName Name of the task
 	 * @param {Function} taskFunction Function which executed the task
-	 * @param {object} taskParams Base parameters for all tasks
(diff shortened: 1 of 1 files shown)


lib/build/TaskRunner.js
```diff
@@ -438,13 +438,12 @@ class TaskRunner {
...
 	 * @param {Function} taskFunction Function which executed the task
-	 * @param {object} taskParams Base parameters for all tasks
 	 * @returns {Promise} Resolves when task has finished
 	 */
-	async _executeTask(taskName, taskFunction, taskParams) {
+	async _executeTask(taskName, taskFunction) {
 		this._log.startTask(taskName);
 		this._taskStart = performance.now();
-		await taskFunction(taskParams, this._log);
+		await taskFunction(this._log);
```
matz3, line 446: The `log` argument is also unused, so we shouldn't pass it here, right?

PR #823, opened 2025-09-12
[INTERNAL] v3: Prevent requesting test packages from npm

d3xter666:
- Do not request packages that are part of the test fixtures from npm, but redirect them to void endpoint.
- Add scoping to the projects, so that they are requested only from `@ui5` npm org that we can control.


codeworrior (approved):
LGTM

PR #835, opened 2025-10-22
[FIX] Revert "[DEPENDENCY] Bump @npmcli/config from 9.0.0 to 10.4.0 #818"

maxreichmann:
This reverts commit a944572.

Previously, `@npmcli/config` required a node version ^20.17.0 || >=22.9.0 which is unsupported by `@ui5/project`.
This PR downgrades it back to ^9.0.0 which requires a node version of ^18.17.0 || >=20.5.0 which fully satisfies our current support.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -2,5 +2,3 @@
 registry=https://registry.npmjs.org/
 lockfile-version=3
 ignore-scripts=true
-# This will fail-fast "npm ci" if deps require node versions unsupported by the repo:
-engine-strict=true
--- a/.npmrc
+++ b/.npmrc
@@ -2,3 +2,5 @@
 registry=https://registry.npmjs.org/
 lockfile-version=3
 ignore-scripts=true
+# This will fail-fast "npm ci" if deps require node versions unsupported by the repo:
+engine-strict=true
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
 			"version": "4.0.6",
 			"license": "Apache-2.0",
 			"dependencies": {
-				"@npmcli/config": "^10.4.2",
+				"@npmcli/config": "^9.0.0",
 				"@ui5/fs": "^4.0.3",
 				"@ui5/logger": "^4.0.2",
 				"ajv": "^6.12.6",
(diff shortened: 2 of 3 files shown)


maxreichmann:
@SAP/ui5-foundation I'd like to hear your opinions regarding `engine-strict=true`.
I've tested it successfully with @npmcli/config ^10.4.0. If it continues to work, we could add it to our other projects (incase there aren't any side effects..)

matz3:
Usually a revert commit should be created with [`git revert`](https://git-scm.com/docs/git-revert) or using the "Revert" button in the pull request. This makes sure that the original commit is referenced and the exact changes are reverted. In this case it will most likely cause merge conflicts in the `package-lock.json`, so it might be indeed easier to manually create a new commit, which is IMHO fine as long it still uses the usual "revert commit" conventions (revert message in commit title / reference to original commit in message body).
Currently there is only a link in the PR description, which is not part of the commit (unless a squash merge is performed to adjust the merged commit message).
Note: A "revert commit" is not a special git commit. It just follows some conventions for the commit message. The `git revert` command basically creates a new commit with the opposite changes of the original commit (which itself is the exact opposite of a `cherry-pick`).

See https://github.com/SAP/ui5-builder/pull/928 as an example.

The original commit message uses the "[DEPENDENCY]" prefix which is included in the changelog as per our config (https://github.com/SAP/ui5-project/blob/v4/.chglog/config.yml#L13). This means we can't easily prevent it from appearing in the changelog. Therefore we should also use a prefix which ends up in the changelog to make clear that this dependency update has been reverted. I would suggest using "[DEPENDENCY]" again here, e.g. `[DEPENDENCY] Revert "Bump @npmcli/config from 9.0.0 to 10.4.0"`

The revert change and also the additional change are rather small, but when reverting changes it is a common best practice to not combine it with other changes to avoid confusion. A revert commit should only revert the changes of the original commit.

Regarding the usage of `engine-strict=true`: I'm okay with adding it to the `.npmrc`, although this would prevent us from using devDependencies with higher engine requirements that are not required for unit testing. In the [UI5/mcp-server](https://github.com/UI5/mcp-server) project this is the case, where `@modelcontextprotocol/inspector` is used for manual testing and it has higher engine requirements. For this reason I would prefer to add this to the CI (by adding `--engine-strict` to the [`npm ci` in the CI workflow](https://github.com/SAP/ui5-builder/blob/a48e474b1bfb9f2f11da44e908341af370711309/.github/workflows/github-ci.yml#L28).

maxreichmann:
@matz3 We do have to revert https://github.com/UI5/cli/commit/f817109e0e02e3b7b832b9c4b7ca2aa5c9951e9f of the monorepo as well, right?

matz3:
> @matz3 We do have to revert [UI5/cli@f817109](https://github.com/UI5/cli/commit/f817109e0e02e3b7b832b9c4b7ca2aa5c9951e9f) of the monorepo as well, right?

I don't think we need to as we will anyways bump our minimum Node.js requirements for the new major release of v5. But until we have done that, it would break a engine-strict check.

matz3:
Actually, I have suggested to use `[DEPENDENCY]` as tag, as this is not related to a issue/bug that has already caused issues for users, for which the `[FIX]` tag is intended to be used.

But as apparently the incompatible bump has just been released as [v4.0.7](https://github.com/SAP/ui5-project/releases/tag/v4.0.7), the `[FIX]` is indeed appropriate, as it is a fix for a problem in an earlier release.

So I'm now approving this PR so that we can produce a follow-up patch to solve the regression.
Please make sure to use **squash and merge** to ensure that the commit message from the PR is used instead of the individual commits (which contain unintended messages).

matz3 (approved):

PR #836, opened 2025-10-23
[INTERNAL] Fail-fast "npm ci" if deps require unsupported node version

maxreichmann:
Follows up https://github.com/SAP/ui5-project/pull/835.

When upgrading a dependency to a newer version, it might require a different node version than before which could be unsupported by the repo. Unfortunately, this could be easily overseen.

This PR adds the flag `--engine-strict` to the dependency installation step during the GH Actions job. Incase such a version requirement change, this flag will now fail-fast the job which should attract attention.

### DIFF
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -25,7 +25,7 @@ jobs:
         node-version: 20.11.0
 
     - name: Install dependencies
-      run: npm ci
+      run: npm ci --engine-strict # --engine-strict is used to fail-fast if deps require node versions unsupported by the repo
 
     - name: Perform checks and tests
       run: npm test


maxreichmann:
Blocked until #835 is merged and this PR is rebased.

matz3:
Please see https://github.com/UI5/cli/blob/v4/docs/Guidelines.md#commit-summary regarding the commit message summary prefix guidelines. `[FIX]` does not seem appropriate here as this does not fix a bug/issue that consumers are facing.

maxreichmann:
> Please see https://github.com/UI5/cli/blob/v4/docs/Guidelines.md#commit-summary regarding the commit message summary prefix guidelines. `[FIX]` does not seem appropriate here as this does not fix a bug/issue that consumers are facing.

I faced this issue when creating this PR but couldn't find any suitable prefix in our conventions: https://github.com/SAP/ui5-project/blob/4636c007b78743ce263d223914832305bda3cb3a/.chglog/config.yml#L9-L14
At the end, it's not a new feature, nor a fix, is not related in any way with improving performance, nor is it a dependency change which should be mentioned in the changelog, nor is it breaking.

It is an internal CI change which does not affect consumers but only contributors. I will change it to "[INTERNAL]" even though we don't specify this in the commit style conventions.

matz3:
The changelog config just lists the tags that are relevant for the changelog, so it is missing there.

The guidelines are mentioning them:
https://github.com/UI5/cli/blob/v4/docs/Guidelines.md#commit-summary
> Use [INTERNAL] for all other changes (e.g. refactorings, documentation, etc.). These changes will not be listed in the changelog.


I notice that you are adding a colon at the end of the tag: `[INTERNAL]:`. What is the reason for that? It is not mentioned in the convention and I haven't seen this in any of our commits so far.

maxreichmann:
Thank you for the clarification and the double-checking 👍 I have removed the colon

matz3 (approved):

PR #859, opened 2026-02-23
[INTERNAL] Bump minimatch to 10.2.2 & tar to 7.5.8

d3xter666:


maxreichmann:
Can you quickly explain why tar is bumped too here?

d3xter666:
> Can you quickly explain why tar is bumped too here?

Basically, the case with the `tar` is generally resolved in the `ui5/cli`, the [`minimatch`](https://github.com/SAP/ui5-cli/pull/832), too.
However, I my arguments to include the fixes also down the chain (`ui5-server` & `ui5-project`) is that eventually someone might use those modules standalone

RandomByte (approved):

PR #883, opened 2026-08-21
[FIX] Prevent npm config from reading UI5 CLI args

matz3:
@npmcli/config reads process.argv by default, which contains our own UI5 CLI args rather than npm args. Pass an empty argv array so these are not misinterpreted as npm configuration.

Down-ported from: https://github.com/UI5/cli/pull/1535


codeworrior (approved):
LGTM

